PR #17, opened 2020-09-16
Possible typo?

timgunter:
Should these "self.rfBuf"s be "self._rxBuf"s?


geeksville:
good find!  definite typos! ;-)

PR #22, opened 2020-09-20
Added "seturl" option to main app

timgunter:
Setting PSK is a little awkward right now from the python API. --info prints out the PSK as a string, but the API needs it fed back as a byte array which as hard to do using --setchan psk ...

Perhaps adding an option to set the PSK and so forth using meshtastic URLs would be easier to use?


geeksville:
looks good to me.  thanks!

PR #24, opened 2020-10-02
Regenerate protobuf docstrings

nils-werner:
When running `bin/regen-docs.sh`, a large amount of docstrings are changed, even though the repo hasn't been touched. I believe this is because somebody had forgotten to regenerate the docs when the protobuf definitions were changed?

Ideally, I believe that eventually the docstrings could be generated from the sourcecode and put on https://readthedocs.org/.

I can set up a docstring pipeline on readthedocs, if you want me to.

In the meantime, as a stop-gap solution, this is a PR that fixes that by simply updating the generated docs.

### DIFF
--- a/docs/meshtastic/mesh_pb2.html
+++ b/docs/meshtastic/mesh_pb2.html
@@ -1713,10 +1713,6 @@ DESCRIPTOR._options = None
 </ul>
 <h3>Class variables</h3>
 <dl>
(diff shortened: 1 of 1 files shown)


geeksville:
thanks!  as for a docstring pipeline on readthedocs.  Can you elaborate?  I know virtually nothing about python best practices for doc distribution.

nils-werner:
Usually you only commit input data to Git, and generate output data in an automated pipeline. In case of Python docs, you would provide all the docs and API references in a machine readable format, and let some automation pipeline generate the HTML output.

The most popular solution is arguably https://readthedocs.org/

I can draft up a PR if you like.

geeksville:
@nils-werner that would be awesome and sounds great!  Yes, I'd love to keep the autogenerated html out of the git repo.

PR #25, opened 2020-10-02
Implement meshtastic.receive.text topic

nils-werner:
Currently, there exist four receive topics:

 - `meshtastic.receive`
 - `meshtastic.receive.position`
 - `meshtastic.receive.user`
 - `meshtastic.receive.data`

I think that receiving text messages is important enough that it warrants its own receive topic

 - `meshtastic.receive.text`

This PR implements that topic.


geeksville:
looks great!

PR #28, opened 2020-10-16
Bug fix for pubsub meshtastic.node.updated not being sent

timgunter:
I think the pub.sendMessage() call for node updates is missing.


geeksville:
thanks!  (And sorry I was afkish for a few days!)

PR #30, opened 2020-10-18
Add factory reset methods to MeshInterface

timgunter:
I've found this helpful for wiping info from the device about nodes I haven't connected to in weeks.

### DIFF
--- a/meshtastic/__init__.py
+++ b/meshtastic/__init__.py
@@ -191,6 +191,12 @@ class MeshInterface:
         self._sendToRadio(toRadio)
         return meshPacket
 
+    def factoryReset(self):
+        t = mesh_pb2.ToRadio()
+        t.set_radio.CopyFrom(self.radioConfig)
+        t.set_radio.preferences.factory_reset = True
+        self._sendToRadio(t)
+
     def writeConfig(self):
         """Write the current (edited) radioConfig to the device"""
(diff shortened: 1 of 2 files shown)


timgunter:
Hm. Something isn't quite right here. When I use my factoryReset() method, I seem to nuke the device's preferences and it crashes periodically.

geeksville:
hmm - let me check.  I might have been lazy and after factory_reset was set to true I just rebooted the device (to ensure all the subsystems rebuild their state)

geeksville:
btw @timgunter would you like me to add you to the org so you can also be a primary contributor/developer/maintainer on the python tool/lib?

timgunter:
> btw @timgunter would you like me to add you to the org so you can also be a primary contributor/developer/maintainer on the python tool/lib?

Sure! I think a lot of the basics are already here and working well. I've been contemplating attempting a pylora implementation of MeshInterface completely independent of the device code. If only I had more free time...

geeksville:
ok great - I'll add you now.  Feel free to (carefully ;-)) add whatever improvements you think make sense for the python api.   Though there is a fair amount of usage now so try to keep backwards compatibility.  If you ever want to be the 'primary' dev (i.e. do pypi releases) for this lib that would be fine with me also ;-).

PR #31, opened 2020-10-25
Added waitForConfig() method to MeshInterface

timgunter:
Convenience method to allow one to block until configuration has been received from the device.


geeksville:
smart!

Issue #39, opened 2020-12-12
Can't send text with meshtastic.py --sendtext


timgunter:
When I run "meshtastic --sendtext some_message", onConnected() closes the interface before the message is sent. Since onConnected is called from a thread managed by pubsub, i'm not sure what the best way to resolve this is. Do we need a way to check if the message has been sent? Or to optionally block until it has been sent? Do we need to add an isBusy threading.Event to MeshInterface?

timgunter:
Perhaps the best thing to do is just set closeNow to False

geeksville:
ooh thanks I think this is a problem with how sent bytes are being treated when closing the serial port.  It seems like they might be getting dropped when the port is closed (even though we previously called serial.flush() to wait for them getting sent.  I think this might mean the problem you saw is a dup of #41 

@timgunter can I ask if you are using windows? (I'm having a hard time reproing this on linux)

geeksville:
I just created a nasty possible fix (see my commit in #41) - would you mind trying it on your machine?

geeksville:
closing as a dup of #41

PR #40, opened 2020-12-13
Don't close now if sending text

timgunter:
This seems to resolve issue #39. Not sure if there aren't similar concerns with other operations or if there is a better way to deal with issues like this.


geeksville:
per other thread I think we have a more fundamental problem when talking to win

Issue #42, opened 2020-12-14
Port nums coming through as strings


timgunter:
Portnums are being stored as strings instead of ints in packet['decoded']['data']['portnum'].

When I register a callback with "meshtastic.receive", and send the device a text message, I get the string 'TEXT_MESSAGE_APP' and not the enum value 1. The portnums_pb2.TEXT_MESSAGE_APP is an int == 1 so comparing the value from the packet to the portnums_pb2.TEXT_MESSAGE_APP returns False.

geeksville:
merged (mostly)

PR #43, opened 2020-12-14
Potential fix for portnums coming through as strings

timgunter:
Is this an acceptable fix for #42? It looks like the intent is for the pubsub topics for apps to use the numeric port number assigned to an app. That is of MY_APP has an enum value of 13, the topic would be "meshtastic.receive.13". I set UNKNOWN_APP as a string to match the way other portnums were being decoded by protobuf. This also should ensure that POSITION_APP and NODEINFO_APP payloads are being handled correctly. This branch also fixes "meshtastic.py --reply" which was not working as a consequence of the string portnum failing the comparison to the int value when choosing how to treat the payload. This ultimately resulted in the payload not being decoded and stored in the dict as "text".

### DIFF
--- a/meshtastic/__init__.py
+++ b/meshtastic/__init__.py
@@ -462,16 +462,20 @@ class MeshInterface:
             # byte array.
             asDict["decoded"]["data"]["payload"] = meshPacket.decoded.data.payload
 
-            # UNKNOWN_APP is the default protobuf pornum value, and therefore if not set it will not be populated at all
+            # UNKNOWN_APP is the default protobuf portnum value, and therefore if not set it will not be populated at all
             # to make API usage easier, set it to prevent confusion
             if not "portnum" in asDict["decoded"]["data"]:
-                asDict["decoded"]["data"]["portnum"] = portnums_pb2.PortNum.UNKNOWN_APP
+                asDict["decoded"]["data"]["portnum"] = portnums_pb2.PortNum.Name(portnums_pb2.PortNum.UNKNOWN_APP)
 
             portnum = asDict["decoded"]["data"]["portnum"]
+
+            if isinstance(portnum, str):
+                portnum = portnums_pb2.PortNum.Value(portnum)
+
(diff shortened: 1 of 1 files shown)


geeksville:
hmm - my intent was that for 'well known' apps the ID would be the string (for readability) and for others it would be the int.  I'll take a look at your PR and add a test case for an unknown portnum and see how that is mapping across.

meshtastic/__init__.py
```diff
@@ -462,16 +462,20 @@ def _handlePacketFromRadio(self, meshPacket):
...
+            # UNKNOWN_APP is the default protobuf portnum value, and therefore if not set it will not be populated at all
             # to make API usage easier, set it to prevent confusion
             if not "portnum" in asDict["decoded"]["data"]:
-                asDict["decoded"]["data"]["portnum"] = portnums_pb2.PortNum.UNKNOWN_APP
+                asDict["decoded"]["data"]["portnum"] = portnums_pb2.PortNum.Name(portnums_pb2.PortNum.UNKNOWN_APP)
 
             portnum = asDict["decoded"]["data"]["portnum"]
+
+            if isinstance(portnum, str):
+                portnum = portnums_pb2.PortNum.Value(portnum)
```
geeksville, line 473: @timgunter what do you think about we take this PR except for these two lines?  Then (it would seem to me - I haven't yet tried this) we would get a string identifier for well-known portnums (for nice readability/logging).  For all other cases the identifier would come through as a number and you'd subscribe for "meshtastic.receive.data.4403" etc...

I think that it is a bit odd to have the IDs come back as strings 'sometimes' but I've kinda been digging it and they are just IDs.

Thoughts?

geeksville (approved):
hmm - I just saw the other bug you helpfully replied to.  Yah - it looks like I busted text message handling.  In the interest of fixing ASAP - I'll merge your change and then add a test for text message and unknown message handling.

PR #50, opened 2021-01-03
Fix MTU - set after mesh interface is up

kareiva:
Fix based on pytap2 docs/examples

https://pypi.org/project/pytap2/


geeksville:
thanks!

Issue #54, opened 2021-02-10
Simplified, tabular output of meshtastic –info 


IZ1IVA:
I was wondering whether it might be a good idea to add a switch to format the output of

`meshtastic --info`

in tabular form **and** reduce data to a minimum when just a quick glance at the mesh status is required.

Output examples:

| user | position | battery | snr |
| ---|---|---|---|
|longName|latitude, longitude, altitude|batteryLevel|snr|

| user | distance | battery | snr |
| ---|---|---|---|
|longName|distance, bearing|batteryLevel|snr|

It could be invoked with

`meshtastic --nodes`

Reference: https://meshtastic.discourse.group/t/simplified-tabular-output-of-meshtastic-info/2468

geeksville:
fixed by @iz1kga - thanks! in https://github.com/meshtastic/Meshtastic-python/pull/56

PR #56, opened 2021-02-12
FIX Issue #54 - implementation of: Simplified, tabular output of meshtastic

iz1kga:


geeksville:
Ooh cool I see you worked it out.  I'll reply when back from lunch.

PR #58, opened 2021-02-16
Nodes print optimizations

iz1kga:
image

Added LastHeard in Human Readable Format
Modified decimal places for printed data
Added units, a number without units is nothing :)

Regards

tanks to @IZ1IVA for testing


geeksville:
Looks great!

PR #59, opened 2021-02-19
Feature requested 20210218

iz1kga:
Removed: Lat, lon, Alt text in position cell
Added: Number, AKA(shortname), Id and Since
Sorted List from newer to older 
Removed self device from list

image


geeksville:
thanks!  somehow I forgot to merge this till now!

PR #74, opened 2021-03-21
Expose API over http server

luandro:
This is very naive implementation (first time trying Python) of how we could expose the API over an http server, in order for radios to communicate over USB with a computer in order to use the web ui.

Ideally we'd have exactly the same endpoints as the web-server that has already been implemented.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -390,6 +390,12 @@ def onConnected(interface):
             closeNow = False
             tunnel.Tunnel(interface, subnet=args.tunnel_net)
 
+        if args.http:
+            from . import http_server
+            # Even if others said we could close, stay open if the user asked for a http server
+            closeNow = False
+            http_server.SimpleHTTPRequestHandler(interface)
+
     except Exception as ex:
         print(f"Exception while handling connection: {ex}")
 
@@ -586,6 +592,9 @@ def initParser():
                         action='store_true', help="Deprecated, use --set is_router true instead")
     parser.add_argument('--unset-router', dest='router',
                         action='store_false', help="Deprecated, use --set is_router false instead")
+
+    parser.add_argument('--http-server',
+                        action='store_false', help="Expose API over http server")
 
     if have_tunnel:
         parser.add_argument('--tunnel',
--- a/meshtastic/http_server.py
+++ b/meshtastic/http_server.py
@@ -1,0 +1,14 @@
+from http.server import HTTPServer, BaseHTTPRequestHandler
+
+class SimpleHTTPRequestHandler(BaseHTTPRequestHandler):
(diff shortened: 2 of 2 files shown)


geeksville:
hmm - see my comment in web chat.  I'm not sure of the usecase of the python API in tern exposing stuff over http.? The device code is already doing that.  Given that do you think there is an extra need for the python code to offer this feature?

luandro:
The device does that through WiFi. The use case I'm thinking of is for doing that same without WiFi, over USB. But probably could be achived via TCP?

geeksville:
but if someone is connecting to the device locally via USB (either android or python), can't they just use the android or python APIs? (which are easier)

luandro:
But how would I use the Python api on a web application?

geeksville:
For a web application, wouldn't you be better off just using meshtastic-js (which can talk directly to the device the same way the python code does)?

luandro:
Meshtastic-js only supports Bluetooth or the http endpoints created by the device using WiFi. When we don't want to use WiFi, and communicate using a usb connection there's currently no way to communicate with the board.

I'm proposing for the python-api to mimic the device generated http api.

Makes sense?

geeksville:

https://meshtastic.discourse.group/t/meshtastic-to-connect-remote-villages-deep-in-the-amazon/2643/14

sachaw:
@luandro I plan to implement webSerial support very soon, which should cover your use case

luandro:
@sachaw I think my usecase might be a bit different, as I won't be acessing the Meshtastic API from the browser of the computer where the LoRa radio is connected to, I'll consuming the API from other devices in the same LAN. Does that make sense?

The easiest path forward for me would be to create a NodeJS API wrapping and exposing the Python API.

sachaw:
By "I'll be consuming the API from other devices in the same LAN", can you elaborate on this because I don't see any situations where this couldn't be achieved how you want it to

sachaw:
What I'm thinking you want to make would be something along the lines of:

Meshtastic device <-> (Node serial <-> exposes arbitrary/proxied JSON API endpoints) <-> some networked client

So basically you have a worker/daemon machine that has a physical connection to a node

luandro:
> Meshtastic device <-> (Node serial <-> exposes arbitrary/proxied JSON API endpoints) <-> some networked client

@sachaw that's exactly what I need. Could Meshtastic.js serial provide for this use case? I thought it was meant only for the browser, but it can also run on NodeJS?

sachaw:
It will, once I add node serial support, which will be soon

luandro:
Ok, that sounds like a plan. Closing this in favor of Meshtastic.js NodeJS serial. Thanks @sachaw and @geeksville for the attention.

PR #79, opened 2021-03-26
Switch to using tabulate to print out node database

timgunter:


geeksville:
VERY NICE!  Thanks for being so proactive and fixing this.  I was kinda dreading it.

PR #80, opened 2021-03-31
Fix position key access in interface.nodes

iz1kga:
Sometimes happens that in meshtastic.Serialinterface().nodes a node hasn't the position key.

In this case meshtastic --nodes will not work because it tries to acces node['position'] key that doesn't exist.

This PR should fix this Behaviour

Regards
Manuel


geeksville:

https://meshtastic.discourse.group/t/1-2-17-alpha-release/3001/15

PR #101, opened 2021-08-13
Updated README and python-cmd-guide

srichs:
The README was updated to remove the deprecated --setchan arg and replace with --ch-set. The python-cmd-guide.md was updated with the current argument list and their usages.


geeksville:
Thank you so much @srichs !

Issue #102, opened 2021-08-18
function should return and not just print out


ChuckNorrison:
https://github.com/meshtastic/Meshtastic-python/blob/037d29e2007efae971d3c05ca721d11f3b6d9976/meshtastic/__init__.py#L230

usage example
```
info = meshtastic_interface.showInfo()
nodes = meshtastic_interface.showNodes()
```

It would be great to return the table to print the information within another python program which calls the function.

jdstroy:
That information should already be present in the API:

```
import meshtastic
ti = meshtastic.TCPInterface("Meshtastic-0123.lan");
ti.nodes["nodeId"]['position'] # is this what you're looking for?
```

ChuckNorrison:
i solved it for me this way, its not beautiful but works

mkinney:
Can we close this issue?

ChuckNorrison:
Works for me :+1:

PR #108, opened 2021-09-06
return and print infos

ChuckNorrison:
return generated informations to the function and not just print out

related to  #102

### DIFF
--- a/meshtastic/__init__.py
+++ b/meshtastic/__init__.py
@@ -170,13 +170,15 @@ class MeshInterface:
 
     def showInfo(self, file=sys.stdout):
         """Show human readable summary about this object"""
-
-        print(
-            f"Owner: {self.getLongName()} ({self.getShortName()})", file=file)
-        print(f"\nMy info: {stripnl(MessageToJson(self.myInfo))}", file=file)
-        print("\nNodes in mesh:", file=file)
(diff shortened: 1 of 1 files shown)


meshtastic/__init__.py
```diff
@@ -227,8 +229,11 @@ def getTimeAgo(ts):
         for i, row in enumerate(rows):
             row['N'] = i+1
 
-        print(tabulate(rows, headers='keys', missingval='N/A',
-                       tablefmt='fancy_grid'), file=file)
```
ChuckNorrison, line 231: maybe someone should check for dependencies to the file=file property, i removed it on this

ChuckNorrison (commented):
cant find any dependencies on this

meshtastic/__init__.py
geeksville: Thanks for the PR.  I don't quite follow this part though, is there a reason you removed the option of emitting the showNodes() output to a destination other than stdout)?

ChuckNorrison: As i changed the prints to assignments, i felt this "file" output was not necessary on this methods, so i tested without them and it works.

geeksville: ok - sounds good!

PR #113, opened 2021-10-21
Fixed order of have_tunnel check

rxt1077:
On Windows (which doesn't have tunnel support) have_tunnel will be false and args.tunnel doesn't exist. When checking to see if this command should continue running the check is evaluated left to right, meaning args.tunnel is checked first giving the error:

AttributeError: 'Namespace' object has no attribute 'tunnel'

Switching the order fixes it.


geeksville:
ooh thanks!

PR #116, opened 2021-11-04
reserved word fix in regen-protos.sh

rxt1077:
It looks like a workaround for the use of reserved words in the protobufs is done by hand (changing None to globals()['None']). This PR adds that step to the regen-protos.sh script.

May be related to Meshtastic-protobufs issue: https://github.com/meshtastic/Meshtastic-protobufs/issues/27

### DIFF
--- a/bin/regen-protos.sh
+++ b/bin/regen-protos.sh
@@ -6,3 +6,6 @@ protoc -I=proto --python_out meshtastic `ls proto/*.proto`
 # workaround for import bug in protoc https://github.com/protocolbuffers/protobuf/issues/1491#issuecomment-690618628
 
 sed -i -E 's/^import.*_pb2/from . \0/' meshtastic/*.py
+
+# automate the current workaround (may be related to Meshtastic-protobufs issue #27 https://github.com/meshtastic/Meshtastic-protobufs/issues/27)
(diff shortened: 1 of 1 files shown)


geeksville:
yay! thanks @rxt1077!

a-f-G-U-C:
> It looks like a workaround for the use of reserved words in the protobufs is done by hand

Careful, please.

The Python "None" error only occurs with old versions of the protobuf compiler, and is fixed automatically in newer versions (so not a manual fix). 
To be clear, there's nothing wrong with this patch, and is indeed very useful as a safety net, but the correct fix is to use a *reasonably recent* version of protoc.

At the moment I use **protoc-3.17** which I selected by trial-and-error to generate files identical to the existing ones (there are subtle template changes from version to version, which show up in the diff).

On this basis, for consistency reasons, it would be nice to specify an "official" protoc version somewhere in the repo, to ensure that all contributors compile the protobufs using the same templates.

rxt1077:
Interesting, I guess we could add a comment at the top of the gen script
saying which version to use? Just like the nanopb comment in the gen script
for meshtastic-device. If there are no objections I'd say the version you
use should be the "official" version.

PR #130, opened 2021-12-01
add cli support for team and position (fixed)

a-f-G-U-C:
Enable the recently added "position fields" and "team" options in the Python CLI tool (updated)


mc-hamster:
thanks for the PR

PR #134, opened 2021-12-03
added smoke tests on one device

### LINKED ISSUE #133: seturl seems to fail silently
If I run this command:

```
meshtastic --seturl "https://www.meshtastic.org/c/#GAMiENTxuzogKQdZ8Lz_q89Oab8qB0RlZmF1bHQ="
```

It appears to work, but in reality, it does not seem to do anything.

The return code is 0. The output is simply "Connected to radio".

Doing a diff of the "--info" before and after shows only one difference "'lastHeard'".

mkinney:
Fix #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,8 +178,11 @@ pip3 install .
 pytest
 ```
 Possible options for testing:
-* for more verbosity, add "-v" or even "-vv" like this: pytest -vv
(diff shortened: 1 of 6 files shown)


geeksville:

https://meshtastic.discourse.group/t/meshtastic-python-1-2-46/4501/1

Issue #146, opened 2021-12-07
Max retransmit error is causing exception as not handled


OverByThere:
Related to #105.
When attempting to get information on a node using `showInfo()` out of range, it returns the following error after a bit of time:
```
ERROR:root:Error while handling message from radio 'admin'
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/meshtastic/__init__.py", line 885, in __reader
    self._handleFromRadio(self._rxBuf[HEADER_LEN:])
  File "/usr/local/lib/python3.7/site-packages/meshtastic/__init__.py", line 566, in _handleFromRadio
    self._handlePacketFromRadio(fromRadio.packet)
  File "/usr/local/lib/python3.7/site-packages/meshtastic/__init__.py", line 704, in _handlePacketFromRadio
    handler.callback(asDict)
  File "/usr/local/lib/python3.7/site-packages/meshtastic/node.py", line 287, in onResponse
    self.radioConfig = p["decoded"]["admin"]["raw"].get_radio_response
KeyError: 'admin'
```

I believe this is because in node.py nothing handles the error in the response:
https://github.com/meshtastic/Meshtastic-python/blob/068b902f12b077daa87cf02cdc876a2e921ff9be/meshtastic/node.py#L270

My crude update to at least report the error and carry on is as follows:
```
def onResponse(p):
    """A closure to handle the response packet"""
    errorFound = False
    if 'routing' in p["decoded"]:
        if p["decoded"]["routing"]['errorReason'] != "NONE":
            errorFound = True
            print(f'Error on response: {p["decoded"]["routing"]["errorReason"]}')
    if errorFound is False:
        self.radioConfig = p["decoded"]["admin"]["raw"].get_radio_response
        logging.debug("Received radio config, now fetching channels...")
        self._timeout.reset()  # We made foreward progress
        self._requestChannel(0)  # now start fetching channels
```

This would then result with the following error output:
`Error on response: MAX_RETRANSMIT`

I cannot propose this as a PR or a full solution as I know its terrible code, but thought I should raise.

mkinney:
I wasn't sure how to replicate the issue.

So, I followed the steps from https://meshtastic.org/docs/software/device/device-remote-admin

one device1
`meshtastic --set factory_reset true`
reboot device
`meshtastic --ch-add admin`
`meshtastic --set is_always_powered true` (optional)
`meshtastic --info` (to get the url)
on a 2nd device:
[...]

PR #158, opened 2021-12-12
add change as suggested by OverByThere

mkinney:
See #146


mkinney:
int and smokewifi passed

smoke1 had one 1 test failed, but I ran that afterwards and it work (so intermittent failure)

geeksville:

https://meshtastic.discourse.group/t/meshtastic-python-1-2-46/4501/1

PR #166, opened 2021-12-16
see if we can validate if we can run the cli from just the build

mkinney:


geeksville:

PR #170, opened 2021-12-17
Add configure dump

### LINKED ISSUE #143: write a "dump-config" type option that would do the opposite of the "--configure" option
It would be nice if we had a way to dump config values to a yaml file.

mkinney:
Fix #143


geeksville:

PR #183, opened 2021-12-29
regen doc with overriding the pdoc _is_public()

mkinney:
Changed this line:

https://github.com/mkinney/pdoc/blob/master/pdoc/__init__.py#L409

to this:

`return not ident_name.startswith("__")`


geeksville:

PR #249, opened 2022-01-30
Device detection

mkinney:
a lot of research on what our devices report in various operating systems (mac, ubuntu, win10)

added a supported_device class and a collection of supported devices

now, we can see if the device(s) are on the supported list - needs more work like T-Echo and DIY


mkinney:
Going to merge this PR and add tests later - want to use this code in the Meshtastic-flasher

PR #302, opened 2022-03-15
Added examples/tcp_gps_example.py

raldi:


raldi:
Is there a more straightforward way to get the local node's position than to look up its nodeNum, then use it to consult the table of all nodes, then pull the position out of there?

mkinney:
Thanks for the additional example.

`make lint` is your friend. :-)

Let me know if you need any help fixing those warnings.

You could change the fixed ip address with one of two options:
a) very simple command line arg, or
b) use hostname "meshtastic.local" - if you have one device that is on your wifi (wifi_ssid and wifi_password) are set, then it should resolve

mkinney:
> Is there a more straightforward way to get the local node's position than to look up its nodeNum, then use it to consult the table of all nodes, then pull the position out of there?

Not sure but @garthvh might have a good idea.

garthvh:
> Is there a more straightforward way to get the local node's position than to look up its nodeNum, then use it to consult the table of all nodes, then pull the position out of there?

That is the right way to do it, positions are on the node

PR #344, opened 2022-06-18
Get and Set general attributes almost working.

caveman99:


caveman99:
Could use a bit of help with the getAttr and setAttr.

PR #349, opened 2022-07-27
Fix crash when trying to set value that doesn't exist

Douile:
The CLI crashes when you get a config key wrong. This fixes that issue.


thebentern:
Great fix! Thanks

PR #353, opened 2022-08-05
Fix lint errors

Douile:
Fix all the current linter errors to make it easier to spot when errors are being added.


thebentern:
Nice!

PR #361, opened 2022-08-16
Fix changing modem_preset from CLI shortcuts

Douile:
Previously the channel shortcuts (e.g. --ch-shortfast) would overrwrite
the entire lora config, this meant you lost other configured things such
as region. This patch changes that so that just modem_preset is
overwritten, and also fixes the call to write config that was missing
an argument.


thebentern:
Nice!

caveman99:
One Hundered! :-)

PR #369, opened 2022-09-06
Fix for --setalt --setlat --setlon

rohanki:
fix for --setalt --setlat and --setlon


thebentern:
Thanks for the fix! Really helpful to get some of these 1.3 config gaps solved in pythhon

rohanki:
That one was a fairly easy one (I think)! I'm still calling myself a python newb, but that compiled OK and tested OK for me...

PR #379, opened 2022-10-08
Catch RoutingApp response for Admin packet

GUVWAF:
When you send an Admin packet to a remote node, it might happen that instead of an AdminApp packet, you get a RoutingApp packet in response due to e.g. a 'NOT_AUTHORIZED' or 'MAX_RETRANSMISSIONS' error. Previously it would then throw a KeyError: 'admin' (as [this issue](https://github.com/meshtastic/Meshtastic-device/issues/1757) also mentions). This will catch the actual routing error and warns the user or retries if the error is 'NONE'.


thebentern:
Thanks. This will be much more informative

PR #386, opened 2022-10-18
python: add QueueStatus support

pavelb-techspark:
This makes Python API parse QueueStatus packages returned by the firmware allowing for TX Queue status control from the phone or PC. This makes it easier to not overwhelm device with packages that are going to be ignored anyways.

Fixes meshtastic/Meshtastic-device#1819

Signed-off-by: Pavel Boldin <pavel.b@techspark.engineering>

### DIFF
--- a/meshtastic/mesh_interface.py
+++ b/meshtastic/mesh_interface.py
@@ -4,6 +4,7 @@ import sys
 import random
 import time
 import logging
+import collections
 from typing import AnyStr
 import threading
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


pavelb-techspark:
I did some testing on this, using ``ping -f -c 100``, out of 100 packets destination node receives ~85 echo requests, with ~70 echo replies successfully received. Most likely this is due to overlapping in time between receiving and sending ends.

On average ping takes ~3 seconds for round trip, given that the queue is empty.
Time difference between last packet sent and last packet received from the source:
```
22:54:28.030570 IP 10.0.105.76 > 10.0.105.88: ICMP echo request, id 77, seq 100, length 64
22:57:06.419293 IP 10.0.105.88 > 10.0.105.76: ICMP echo reply, id 77, seq 100, length 64
```

Same packet on the destination:
```
22:57:02.296597 IP 10.0.105.76 > 10.0.105.88: ICMP echo request, id 77, seq 100, length 64
22:57:02.296621 IP 10.0.105.88 > 10.0.105.76: ICMP echo reply, id 77, seq 100, length 64
```

PR #395, opened 2022-11-02
Change how admin packets to remote nodes are handled

GUVWAF:
To limit the amount of packets sent for an Admin message, do not ask for an Ack if you ask for a Response. Since the device considers a Response as an Ack as well, the behaviour doesn't change.


GUVWAF:
For admin packets that don’t need to know the configs or channels of the remote node (reboot, shutdown, factory-reset, reset-nodedb, set-canned-message, set-owner), now it doesn’t request those, but waits until it gets an (implicit) ACK or NAK to inform you whether the transmission was successful.

thebentern:
This looks great!

PR #428, opened 2023-03-19
Get and set methods for remote node using one admin message

### LINKED ISSUE #423: remote admin `--info` returns empty "Preferences" and "Module preferences"
```
(venv) meshtastic-python % meshtastic --dest '!remote_id' --info
Connected to radio


Requesting channel 0 info from remote node (this could take a while)
Requesting channel 1 info from remote node (this could take a while)
Requesting channel 2 info from remote node (this could take a while)
Requesting channel 3 info from remote node (this could take a while)
Requesting channel 4 info from remote node (this could take a while)
Requesting channel 5 info from remote node (this could take a while)
Preferences: {}

Module preferences: {}

Channels:
  PRIMARY psk=secret { "psk": "blablabla", "name": "ChannelName" }

Primary channel URL: https://meshtastic.org/e/#blablabla
```

GUVWAF:
Now the commands `--get` and `--set` work for remote nodes by using only one admin message per field and it doesn’t request all channels. When setting, it first requests the current config of the field, then sends an updated admin message, after which it waits for an ACK from the remote node.

This partly fixes #423. When trying `--info` for a remote node, it prints that you need to use `--get` with a specific field. Also, it prints that `--nodes` for a remote node is not supported (currently it was showing the list of the connected node).

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -53,11 +53,11 @@ def onConnection(interface, topic=pub.AUTO_TOPIC): # pylint: disable=W0613
     print(f"Connection changed: {topic.getName()}")
 
 
-def getPref(interface, dest, comp_name):
+def getPref(node, comp_name):
     """Get a channel or preferences value"""
 
     name = splitCompoundName(comp_name)
-    fullConfig = name[0] == name[1] # We want the full config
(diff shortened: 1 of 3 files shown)


garthvh (approved):
Looks good

PR #429, opened 2023-03-19
Convert sections to snake_case when configuring

### LINKED ISSUE #426: [Bug]: canned_message and external_notification named incorrectly on export
When exporting the config file, the generated yaml file incorrectly names the external_notification and canned_message modules as `cannedMessage` and `externalNotification` and as such these sections will not import correctly. `Error: No valid config with name cannedMessage` or `Error: No valid config with name externalNotification`. 

I can confirm if you make these changes to the yaml file that is exported, when you go to import it'll import these configs without issues. 

`cannedMessage` --> `canned_message`
`externalNotifcation` --> `external_notification`

GUVWAF:
Fixes #426. Since everything is CamelCase in the export, instead of changing this, now `--configure` converts it correctly.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -459,15 +459,15 @@ def onConnected(interface):
                     localConfig = interface.getNode(args.dest).localConfig
                     for section in configuration['config']:
                         for pref in configuration['config'][section]:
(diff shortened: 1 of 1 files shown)


garthvh (approved):
Good idea!

PR #446, opened 2023-04-14
Update Supported Device List

rcarteraz:
Not sure if this matters or not but I noticed the list of devices in supported_device.py was seriously outdated. I've done my best to format as required based on how others are and I think I've got it mostly right. The usb vid/pid for the t3s3 and station g1 and nano g1 explorer devices I'm not sure if are accurate or not as I don't have them. Most of the devices in a family seem to share these so I'm assuming that's the case here but maybe someone with them can double check? Leaving this as a draft for input for now as it's kind of out of my wheelbase.

### DIFF
--- a/meshtastic/supported_device.py
+++ b/meshtastic/supported_device.py
@@ -275,6 +275,7 @@ supported_devices = [
     heltec_v2_0,
     heltec_v2_1,
     heltec_v3,
+    heltec_wsl_v3,
     meshtastic_diy_v1,
     techo_1,
     rak4631_5005,


meshtastic/supported_device.py
```diff
@@ -205,24 +250,39 @@ def __init__(
...
     heltec_v2_0,
     heltec_v2_1,
+    heltec_v3,
+    heltec_wsl_v3,
     meshtastic_diy_v1,
     techo_1,
     rak4631_5005,
     rak4631_5005_epaper,
     rak4631_19003,
     rak11200,
```
caveman99, line 284: The RAK targets seem to be out of sync too. We dropped the distinction 5005/19003 a while ago.

caveman99 (changes requested):

meshtastic/supported_device.py
OpenSource-For-Freedom: Do we need to update this list?

OpenSource-For-Freedom:
(This is Tank on Discord) Ill compile a list for approval prior to pulling this file if thats ok. I found a large list of new devices.

PR #462, opened 2023-10-14
Categorized errors, updated return & conditional statements.

ProgramR4732:
Made the following changes:

-> Categorized Exception error into sub error categories.
-> Replaced a nested if statement.
-> Replaced explicit 'return None' to 'return' as both mean the same.


jp-bennett (approved):

thebentern:
Please sign the CLA

PR #505, opened 2024-03-14
Show unknown nodes in a fashion similar to the web UI. Fixes #504

### LINKED ISSUE #504: [Improvement] Show unknown nodes in some fashion in --nodes output
When nodes have been seen (via ack or possibly other ways) but don't have complete information (maybe if they only have an ID -- unsure), the CLI doesn't seem to show them at all. This matches Android (meshtastic/Meshtastic-Android#912) but does not match the web client that shows `UNK: <integer node ID>` nor iOS which shows `Meshtastic XXXX` for those nodes. My comments over on that issue talk about why I think it'd be useful to see those nodes in the list even given very limited information about them.

ianmcorvidae:
Similar to how the web UI shows "UNK: <integer node ID>", this will show that under "User", and the hex ID that can be computed from that under "ID". The change to using `self.nodesByNum` was necessary to pull in those, since `self.nodes` only has nodes that have users, which excludes nodes in this particular state.

### DIFF
--- a/meshtastic/mesh_interface.py
+++ b/meshtastic/mesh_interface.py
@@ -152,13 +152,13 @@ class MeshInterface:
             )
 
         rows = []
-        if self.nodes:
+        if self.nodesByNum:
             logging.debug(f"self.nodes:{self.nodes}")
-            for node in self.nodes.values():
+            for node in self.nodesByNum.values():
                 if not includeSelf and node["num"] == self.localNode.nodeNum:
                     continue
 
-                row = {"N": 0}
+                row = {"N": 0, "User": f"UNK: {node['num']}", "ID": f"!{node['num']:x}"}
 
                 user = node.get("user")
                 if user:


jp-bennett (approved):
I don't have any unknown nodes to confirm, but the node list looks fine otherwise.

ianmcorvidae:
Merged already (thanks!), but I should have included what it looks like via my own couple of 'em, so I'll still drop that here:

```
│  14 │ UNK: 2889340535                  │ !ac37d677 │ N/A   │ N/A        │ N/A         │ N/A        │ N/A       │ N/A             │ N/A            │ 13.25 dB  │       N/A │ 2024-03-02 11:42:26 │ 1 week ago     │
├─────┼──────────────────────────────────┼───────────┼───────┼────────────┼─────────────┼────────────┼───────────┼─────────────────┼────────────────┼───────────┼───────────┼─────────────────────┼────────────────┤
│  15 │ UNK: 2208397066                  │ !83a1770a │ N/A   │ N/A        │ N/A         │ N/A        │ 79%       │ 8.15%           │ 0.50%          │ 10.75 dB  │       N/A │ 2024-03-02 11:41:37 │ 1 week ago     │
```

PR #516, opened 2024-03-19
use importlib.metadata and packaging.version instead of pkg_resources

ianmcorvidae:
This will, hopefully, fix or at least improve on the CI situation. I've been having a bit of trouble running a local runner for github actions and getting it to cooperate, but it seems to at least be getting further. We'll see, maybe?


ianmcorvidae:
okay, this may need more work, heh. I'll keep plugging at it a bit

PR #521, opened 2024-03-21
feat: Send ping with a channel index

renanbastos93:
Now we can send ping messages with a channel index. 
Additionally, a new unit test has been added to simulate this functionality, and other tests have been fixed.

e.g
```sh
$ meshtastic --sendping --ch-index 1 
# output:
# Connected to radio
# Sending ping message to ^all
```

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -395,9 +395,6 @@ def onConnected(interface):
                 )
 
         if args.sendping:
-            channelIndex = 0
-            if args.ch_index is not None:
-                channelIndex = int(args.ch_index)
             payload = str.encode("test string")
             print(f"Sending ping message to {args.dest}")
             interface.sendData(
@@ -406,7 +403,7 @@ def onConnected(interface):
                 portNum=portnums_pb2.PortNum.REPLY_APP,
                 wantAck=True,
                 wantResponse=True,
-                channelIndex=channelIndex,
+                channelIndex=our_globals.get_channel_index() or 0,
             )
 
         if args.traceroute:
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -395,6 +395,9 @@ def onConnected(interface):
                 )
 
         if args.sendping:
+            channelIndex = 0
+            if args.ch_index is not None:
+                channelIndex = int(args.ch_index)
             payload = str.encode("test string")
             print(f"Sending ping message to {args.dest}")
             interface.sendData(
@@ -403,6 +406,7 @@ def onConnected(interface):
                 portNum=portnums_pb2.PortNum.REPLY_APP,
                 wantAck=True,
                 wantResponse=True,
+                channelIndex=channelIndex,
             )
 
         if args.traceroute:
--- a/meshtastic/tests/test_main.py
+++ b/meshtastic/tests/test_main.py
@@ -722,9 +722,34 @@ def test_main_sendping(capsys):
 
     iface = MagicMock(autospec=SerialInterface)
 
-    def mock_sendData(payload, dest, portNum, wantAck, wantResponse):
+    def mock_sendData(payload, dest, portNum, wantAck, wantResponse, channelIndex):
         print("inside mocked sendData")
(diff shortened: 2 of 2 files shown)


meshtastic/__main__.py
```diff
@@ -395,6 +395,9 @@ def onConnected(interface):
                 )
 
         if args.sendping:
+            channelIndex = 0
+            if args.ch_index is not None:
+                channelIndex = int(args.ch_index)
```
ianmcorvidae, line 400: I see why you did this since it's copying what `sendtext` does below, but you probably shouldn't need to look at `args.ch_index` again here. Within `common()` we pull the channel index into our_globals. So I think this should be `channelIndex = our_globals.get_channel_index() or 0`

ianmcorvidae (changes requested):
I had one small improvement to suggest. I'm assuming these tests as you've written them pass; there's a lot of work to improve the tests that's needed (I have a branch in progress that at least gets things back to passing), so I imagine there might need to be more changes there, but that shouldn't block this PR as long as what you've added/changed still works, I think.

renanbastos93:
> I had one small improvement to suggest. I'm assuming these tests as you've written them pass; there's a lot of work to improve the tests that's needed (I have a branch in progress that at least gets things back to passing), so I imagine there might need to be more changes there, but that shouldn't block this PR as long as what you've added/changed still works, I think.

Yes, it's working. I've conducted numerous tests locally

meshtastic/__main__.py
renanbastos93: Of course, I got it, I have been changing that.

ianmcorvidae (approved):
Looks good to me! We'll see if anyone with write access has any other thoughts before merging, but I at least think it's ready to go.

garthvh:
The reply app has been deprecated in recent firmware, it was designed to be a developer starting point so it is commented out and not a active port anymore in firmware releases.

renanbastos93:
> The reply app has been deprecated in recent firmware, it was designed to be a developer starting point so it is commented out and not a active port anymore in firmware releases.

I apologize for the confusion. Could you please clarify if this CLI has been deprecated? If so, which CLI should I use instead?

ianmcorvidae:
Oh right, I forgot about that completely. My feeling is that we can merge this, but also add a warning to `--sendping` that REPLY_APP is deprecated, and remove it from this library in a couple point releases.

@renanbastos93:  the CLI itself isn't deprecated, but the ping/auto-reply functionality is in the process of being removed from the meshtastic firmware, so the messages that this sends won't get replies from nodes with newer firmware, and so won't be supported in any clients going forward. So users of the ping will need to find alternatives for their use cases regardless.

renanbastos93:
> Oh right, I forgot about that completely. My feeling is that we can merge this, but also add a warning to `--sendping` that REPLY_APP is deprecated, and remove it from this library in a couple point releases.
> 
> @renanbastos93: the CLI itself isn't deprecated, but the ping/auto-reply functionality is in the process of being removed from the meshtastic firmware, so the messages that this sends won't get replies from nodes with newer firmware, and so won't be supported in any clients going forward. So users of the ping will need to find alternatives for their use cases regardless.

Oh, I understand now. So if I need to undergo a treatment involving ping, I'll have to explore alternative options.

garthvh:
It got pulled as it was pretty spammy and was incomplete, not sure if it should be in the CLI, pretty quickly it won't work for most users.

renanbastos93:
> It got pulled as it was pretty spammy and was incomplete, not sure if it should be in the CLI, pretty quickly it won't work for most users.

It's okay, no worries. Do you believe we should close this PR?

renanbastos93:
@garthvh anything else?

ianmcorvidae:
Sorry not to get back to you. I do think at this point we should close this PR given this functionality is being removed from firmware. I'll get something together separately to remove `--sendping`entirely.

renanbastos93:
> Sorry not to get back to you. I do think at this point we should close this PR given this functionality is being removed from firmware. I'll get something together separately to remove `--sendping`entirely.

of course dude, see ya

PR #524, opened 2024-03-24
Fix up or comment out broken tests, to get CI (hopefully) happy

ianmcorvidae:
There's a couple other things in here as well -- fixing up the tests did find at least one actual bug.

Not super happy with how many of these I commented out, but it gets us back to a point where we can expect new PRs to maintain tests passing, which I think is valuable anyway.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -160,8 +160,8 @@ def setPref(config, comp_name, valStr) -> bool:
     val = meshtastic.util.fromStr(valStr)
     logging.debug(f"valStr:{valStr} val:{val}")
 
-    if snake_name == "psk" and len(valStr) < 8:
-        print(f"Warning: wifi.psk must be 8 or more characters.")
+    if snake_name == "wifi_psk" and len(valStr) < 8:
+        print(f"Warning: network.wifi_psk must be 8 or more characters.")
         return False
 
     enumType = pref.enum_type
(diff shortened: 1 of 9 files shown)


meshtastic/__main__.py
```diff
@@ -160,8 +160,8 @@ def setPref(config, comp_name, valStr) -> bool:
     val = meshtastic.util.fromStr(valStr)
     logging.debug(f"valStr:{valStr} val:{val}")
 
-    if snake_name == "psk" and len(valStr) < 8:
-        print(f"Warning: wifi.psk must be 8 or more characters.")
+    if snake_name == "wifi_psk" and len(valStr) < 8:
+        print(f"Warning: network.wifi_psk must be 8 or more characters.")
```
ianmcorvidae, line 164: This bit hadn't been updated for the new config format, so that's a bug fixed at least.

meshtastic/__main__.py
```diff
@@ -638,6 +638,11 @@ def onConnected(interface):
 
         def setSimpleConfig(modem_preset):
             """Set one of the simple modem_config"""
+            channelIndex = our_globals.get_channel_index()
+            if channelIndex is not None and channelIndex > 0:
+                meshtastic.util.our_exit(
+                    "Warning: Cannot set modem preset for non-primary channel", 1
+                )
```
ianmcorvidae, line 645: We had a test expecting to enforce this, but weren't doing so. So if channel index is passed along with a modem preset, we expect it to be 0 now.

meshtastic/mesh_interface.py
```diff
@@ -129,7 +129,6 @@ def showInfo(self, file=sys.stdout):  # pylint: disable=W0613
 
                 # use id as dictionary key for correct json format in list of nodes
                 nodeid = n2["user"]["id"]
-                n2["user"].pop("id")
```
ianmcorvidae, line 132: I'm not sure why we did this other than to clean up the display, but it meant if this got called and later something else needed the ID, it would fail.

ianmcorvidae (commented):
(a few of the not immediately test-related fixes noted here)

ianmcorvidae:
(the failures now are about uploading to codecov. will try to look into that soon)

thebentern (approved):

thebentern (approved):

PR #526, opened 2024-03-26
Add missing portnums to the dict for automatic decoding as protocol buffers/text

ianmcorvidae:
This is the dict that lets the code know "if you get something on this portnum, decode/process it this way".

Range Test and Detection Sensor are both noted as text in portnums.proto. The others are just set up to decode as the appropriate protobufs.

@pdxlocations , I requested a review from you because https://github.com/pdxlocations/Meshtastic-Python-Examples/blob/main/print-packets.py uses most of these and as such I figured you might be well-placed to make sure I didn't screw something up. And you might want to take advantage of the update, too!

### DIFF
--- a/meshtastic/__init__.py
+++ b/meshtastic/__init__.py
@@ -84,8 +84,11 @@ from meshtastic import (
     channel_pb2,
     config_pb2,
     mesh_pb2,
+    mqtt_pb2,
+    paxcount_pb2,
     portnums_pb2,
     remote_hardware_pb2,
+    storeforward_pb2,
     telemetry_pb2,
     util,
 )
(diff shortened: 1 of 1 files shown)


meshtastic/__init__.py
```diff
@@ -190,6 +193,13 @@ def _receiveInfoUpdate(iface, asDict):
     portnums_pb2.PortNum.TEXT_MESSAGE_APP: KnownProtocol(
         "text", onReceive=_onTextReceive
     ),
+    portnums_pb2.PortNum.RANGE_TEST_APP: KnownProtocol(
+        "rangetest", onReceive=_onTextReceive
+    ),
+    portnums_pb2.PortNum.DETECTION_SENSOR_APP: KnownProtocol(
+        "rangetest", onReceive=_onTextReceive
```
pdxlocations, line 200: probably:
"detectionsensor", onReceive=_onTextReceive

pdxlocations (commented):
I'm slowing learning how things work around hear.  Happy to review but don't trust anything I say. :-)

meshtastic/__init__.py
ianmcorvidae: That would make more sense, huh. Copy-paste error on my part there, haha

thebentern (approved):

PR #528, opened 2024-03-30
Protobuf type interfaces for type-checking & undoing protobufs hack

ianmcorvidae:
There's a lot of changes here but the majority of them are generated code. The big thing is that we're not rewriting the protobufs to undo the change to namespaced stuff, and we're adding in a plugin that generates `.pyi` files that denote types for the protobufs. There's also a few other miscellaneous changes that mostly boil down to me making mypy happy and fixing the one bona fide bug I found. I'll leave a review pointing out the non-generated changes.


bin/regen-protobufs.sh
```diff
@@ -4,8 +4,8 @@
 #gsed -i 's/import "\//import ".\//g' ./protobufs/meshtastic/*
 #gsed -i 's/package meshtastic;//g' ./protobufs/meshtastic/*
 
-./nanopb-0.4.7/generator-bin/protoc -I=protobufs --python_out ./ ./protobufs/meshtastic/*.proto
-./nanopb-0.4.7/generator-bin/protoc -I=protobufs --python_out ./meshtastic/ ./protobufs/nanopb.proto
+./nanopb-0.4.7/generator-bin/protoc -I=protobufs --python_out ./ --mypy_out ./ ./protobufs/meshtastic/*.proto
+./nanopb-0.4.7/generator-bin/protoc -I=protobufs --python_out ./meshtastic/ --mypy_out ./meshtastic/ ./protobufs/nanopb.proto
```
ianmcorvidae, line 8: This is what gets the `pyi` files to be generated.

meshtastic/__main__.py
```diff
@@ -9,16 +9,15 @@
...
 from google.protobuf.json_format import MessageToDict
-from pubsub import pub
+from pubsub import pub # type: ignore[import-untyped]
 
 import meshtastic.test
 import meshtastic.util
-from meshtastic import channel_pb2, config_pb2, portnums_pb2, remote_hardware
+from meshtastic import channel_pb2, config_pb2, portnums_pb2, remote_hardware, BROADCAST_ADDR
 from meshtastic.version import get_active_version
-from meshtastic.__init__ import BROADCAST_ADDR
```
ianmcorvidae, line 21: Removing this import from `meshtastic.__init__` was something mypy needed to be happy, and is done a few different places. Otherwise it would complain about the same file being imported both as `meshtastic` and as `meshtastic.__init__`. Plus importing from `__init__` is a little bit funky anyway, pythonically speaking.

meshtastic/mesh_interface.py
```diff
@@ -9,17 +9,18 @@
...
 import time
 from datetime import datetime
-from typing import AnyStr
 
 import google.protobuf.json_format
-import timeago
+import timeago # type: ignore[import-untyped]
 from google.protobuf.json_format import MessageToJson
-from pubsub import pub
+from pubsub import pub # type: ignore[import-untyped]
```
ianmcorvidae, line 16: These sorts of `ignore[import-untyped]` silence mypy a bit when using libraries that don't have type hints (or at least, aren't marked as such) nor type stubs.

meshtastic/mesh_interface.py
```diff
@@ -239,7 +240,7 @@ def getNode(self, nodeId, requestChannels=True):
 
     def sendText(
         self,
-        text: AnyStr,
```
ianmcorvidae, line 242: While this was marked as `AnyStr` (which is basically "either string or bytes, but not a mix of the two"), we treated it afterwards as always being a string. mypy complained about that and I think it's right to have this be a plain str instead.

meshtastic/node.py
```diff
@@ -123,7 +123,7 @@ def requestConfig(self, configType):
             print("Requesting current config from remote node (this can take a while).")
 
         msgIndex = configType.index
-        if configType.containing_type.full_name == "LocalConfig":
+        if configType.containing_type.full_name in ("meshtastic.LocalConfig", "LocalConfig"):
```
ianmcorvidae, line 126: This is the one actual thing that just plain didn't work. The namespacing changed the full name, so trying to fetch things in LocalConfig actually fetched from ModuleConfig.

meshtastic/tcp_interface.py
```diff
@@ -30,10 +30,10 @@ def __init__(
         self.portNumber = portNumber
 
         if connectNow:
-            logging.debug(f"Connecting to {hostname}")
+            logging.debug(f"Connecting to {hostname}") # type: ignore[str-bytes-safe]
             server_address = (hostname, portNumber)
             sock = socket.create_connection(server_address)
-            self.socket = sock
+            self.socket: Optional[socket.socket] = sock
```
ianmcorvidae, line 36: This made mypy happier since this could be set to None as well as being an actual socket.

requirements.txt
```diff
@@ -20,3 +20,10 @@ pdoc3
 pypubsub
 bleak
 packaging
+mypy
+mypy-protobuf
+types-protobuf
+types-tabulate
+types-requests
+types-setuptools
+types-PyYAML
```
ianmcorvidae, line 29: Probably self-explanatory, but this adds mypy, the mypy-protobuf plugin that generates the `.pyi` files, and type stubs for what existed.

ianmcorvidae (commented):
I haven't yet set up mypy to run in CI, which is sort of the outstanding piece of this.

ianmcorvidae:
Alright, I'm gonna take the risk and merge this in and maybe/hopefully start adding some more types. Here's hoping it doesn't break everything!

PR #544, opened 2024-04-10
Refactor to avoid the use of a special global object.

holdenweb:
The global object formerly used as a singleton is now replaced by direct use of the namespace of the globals module. This eliminates the redundant getters and setters, and both clarifies and simplifies the code for future maintainers.

Note that the globals module name conflicts (harmlessly at present) with a Python built-in function. A future commit could rename it `config` to remove this clash and better represent its intended purpose.

The impact of this change may suggest other fairly simple improvements of a similar nature to improve maintainability and testability.


holdenweb:
Ah, I see I borked the code. For some reason I find pylint unusable locally - it finds far more errors than the CI chain is doing. I'll fix the errors I see reported in the CI output and amend the PR. Sorry about the hassle.

ianmcorvidae:
Will try to take a look at this later, but I'll at least set off a CI run in case there's any other issues there that might come up.

holdenweb:
I see that the test coverage going down has raised a flag. I suspect this may be because I deleted a couple of tests that became redundant with the removal of a global object. I trust this slight negative is outweighed by the positive impact of the changes. The code base has actually shrunk by almost 100 lines.

ianmcorvidae:
Yeah, I wouldn't block it on the basis of just that. There's a lot of improvements to be made to the tests, heh.

holdenweb:
Thanks for merging. I'm currently looking at a whole range of potential refactorings, but before submitting further pull requests would appreciate the change to talk about them first, if you have chance. I don't want to start proposing changes that don't align with your future direction.

holdenweb:
> ... and it'll be nice if we can move away from needing anything to do with globals outright eventually ...
Yes indeed. Ideally each interface would be self-contained and allow us to build sensible systems with multiple radios. Regards this as a first simolification step along the way.

ianmcorvidae:
I'm happy to chat about stuff, for sure! I don't know if you're on the Meshtastic discord but I hang around there a decent amount (I'm `m_ia_n` there), or of course here/email. In a lot of ways I'm not super opinionated about the future direction the library and CLI go -- mostly, I want to avoid breaking people's existing scripts & programs, keep up with new features, and improve things as far as usability, maintainability, and such. Which is to say I'm open to hearing any ideas and welcome any help!

PR #553, opened 2024-04-21
Disallow --dest in remaining spots that don't actually use/allow it

ianmcorvidae:
(at least assuming I didn't miss any)


ianmcorvidae:
I'll be merging this in a second assuming CI is happy (enough), which will also merge #551 and #552

PR #554, opened 2024-04-22
Add missing StoreForward config from radio

GUVWAF:
`--info` didn't have the StoreForward config and `--get store_forward.enabled` would always return `false` due to missing config handler.


ianmcorvidae:
Ah, good catch. Still learning the various spots that need manual updating like this, heh. Thanks for the patch!

PR #564, opened 2024-05-03
Fix typo in __main__.py

jhollowe:
fix small typo in help test for `--request-position`


ianmcorvidae:
Cool, thanks for this. Must have gotten my eyes crossed and missed it when I added that!

Issue #569, opened 2024-05-09
Discussion: machine-readable output


ianmcorvidae:
Breaking this out into a new issue for discussion from a couple other spots.

We should establish a format and standard way of accessing machine-readable output, most likely JSON, and then implement it in the CLI. Once we have an initial standard, other output options such as tabular formats can be more easily defined relative to a JSON format.

My preferred format would diverge as little as possible from things as defined in [the protobufs](https://buf.build/meshtastic/protobufs/docs/main:meshtastic) for the sake of easier forward compatibility; if a field gets added somewhere, updating the protocol buffer definitions should as often as possible be all that's needed.

I think that there's probably several good ways to go about this, and the decisions may be closely tied with any reorganization of the CLI. Broadly, I think we might have two options:
 * a single format that's able to express anything relevant within the same structure, which different commands and options would populate different parts of (for read operations) or only read/consider specific parts of (for write/update operations)
 * a number of somewhat more command-specific formats not intended to be directly merged with one another but instead relying on the context of the command
 
To me, the first of these would be more consistent but more challenging to define fully and less well-suited for streaming data, while the second would be a little easier to create and better-suited for streaming but require more work from users in making their scripts understand the context of the data.
 
I don't have any sort of exclusive claim on ideas and there could easily be good options I haven't thought of! So please share your thoughts if you're a user of the CLI or might want to be a user of machine-readable outputs from the CLI.

Previously: #475, #566

holdenweb:
Sounds like we're looking for a generic "Protobuf to JSON" function. Let me think about that.

ianmcorvidae:
Yeah -- it's more of a question of how to put the outputs of that together. We already use MessageToJson (and the related MessageToDict) a lot, but there isn't something in the protobufs that defines a completely top-level format like my first option above, and we'd need to decide which things to return where for the second option.

holdenweb:
Each protobuf has a unique name, right? So how about a dict keyed by protobuf name, whose values were a list of zero or more dicts each representing a protobuf's contents. If the list associated with a key were empty then that key could be omitted.

There may be other data we wish to communicate. We can include that as (an)other key(s) in the same dict. Would that work?

ianmcorvidae:
[...]

PR #577, opened 2024-05-16
--json Outputs JSON objects for --nodes, --request-telemetry/position

aikomastboom:
In a attempt to make the output of meshtastic a bit more machine parsing friendly (eg by [nushell](https://github.com/nushell/)) I propose the addition of '--json' to switch the output to something easily parsable when piped on the commandline.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -253,7 +253,7 @@ def onConnected(interface):
         args = mt_config.args
 
         # do not print this line if we are exporting the config
-        if not args.export_config:
+        if not (args.export_config or args.json):
             print("Connected to radio")
 
         if args.setlat or args.setlon or args.setalt:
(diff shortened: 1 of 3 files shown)


holdenweb:
Am I alone in thinking that any output produced in the presence of a `--json` option should actually be readable as a JSON document? I'm quite happy for other formats to exist, but please let's not mangle existing standards. Transforming JSON is a straightforward task, for which many support libraries exist. Parsing invented almost-JSON formats isn't.

ianmcorvidae:
It appears that these formats are, in fact, JSON (everything is produced via `json.dumps`) and could be treated as such.

However, they do look fairly ad-hoc and I think we need a consistent approach, as we're discussing over in #569 -- the discussed plan there so far has largely been to do as little modification of protobufs as we can other than turning them to JSON, and to identify things by the (lowercased) name of the protobuf. This will probably require a bit of internal refactoring for some things (e.g. I don't believe right now we're holding onto a copy of the actual `NodeInfo` protobufs for nodes, but packing into our own structure instead), but would be much more easily extensible to any other commands and straightforwardly documented in terms of the protobuf definitions.

aikomastboom:
I totally agree.. it would be preferred to eliminate conversions back and forth into JSON like this. Refactoring the whole codebase to facilitate a clear separation of concern / interface was not my intention. I guess It's best to close this PR and keep the effort to myself while a proper solution is drafted in [Discussion: machine-readable output](https://github.com/meshtastic/python/issues/569)

PR #580, opened 2024-05-24
Add "wait to disconnect" argument

todd-herbert:
This pull request adds a `--wait-to-disconnect [seconds]` argument. When set, the code sleeps before closing the interface. If used without specifying [seconds], a default value of 5 seconds is used. 

This helps with hardware which resets when the serial connection is closed. When setting config, these devices can sometimes reset before NodeDB can be saved.


ianmcorvidae:
This looks good to me! Sorry for the delay in looking at it, seems like a good addition.

Issue #568, opened 2024-05-08
Error on meshtastic --set lora.ignore_incoming xxxxxx


868meshbot:
Trying to use the python client, to manage ignore_incoming for testing purposes and i get the below error. Included my protoc and pip versions on osx.  Not sure where I've gone wrong? any pointer? thanks

```
meshtastic --set lora.ignore_incoming xxxxxx
Connected to radio
Adding 'xxxxx' to the ignore_incoming list
Aborting due to: 'google._upb._message.Descriptor' object has no attribute 'ignore_incoming'
```
or 
```
Aborting due to: 'Descriptor' object has no attribute 'ignore_incoming'
```

```
protoc --version
libprotoc 26.1

pip freeze|grep proto
mypy-protobuf==3.6.0
protobuf==5.26.1
types-protobuf==5.26.0.20240422

```
EDIT: Updated protobufs manually recompiled and reinstalled v2.3.9 of python 
```
DEBUG file:__main__.py setPref line:163 snake_name:ignore_incoming
DEBUG file:__main__.py setPref line:164 camel_name:ignoreIncoming
DEBUG file:__main__.py setPref line:185 valStr:xxxxxxxx val:xxxxxxxx
Adding 'xxxxxxxx' to the ignore_incoming list
Aborting due to: 'Descriptor' object has no attribute 'ignore_incoming'
DEBUG file:serial_interface.py close line:74 Closing Serial stream
DEBUG file:stream_interface.py close line:119 Closing stream
```

868meshbot:
Fixed it and pushed a pull request

868meshbot:
Whoops hit comment instead of close see https://github.com/meshtastic/python/pull/581

PR #581, opened 2024-05-25
Fix the ignore_incoming management BUG 568

868meshbot:
Simple patch to fix the ignore_incoming management aka https://github.com/meshtastic/python/issues/568


ianmcorvidae:
I think this looks good, I'll go ahead and merge it. Thanks for the fix!

PR #582, opened 2024-05-28
Clarify licensing

FedericoCeratto:
Hello and thank you for developing Meshtastic!
Unclear licensing make it difficult to package and distribute software without causing legal risks for end users.
This PR is mean to clarify licensing in compliance with https://reuse.software/
The license was changed on 1 Nov 2022 in https://github.com/meshtastic/python/pull/392

The second commit annotates files with licensing information - if you find it too invasive I can remove it.

Thanks!


README.md
```diff
@@ -5,6 +5,7 @@
 [![CI](https://img.shields.io/github/actions/workflow/status/meshtastic/python/ci.yml?branch=master&label=actions&logo=github&color=yellow)](https://github.com/meshtastic/python/actions/workflows/ci.yml)
 [![CLA assistant](https://cla-assistant.io/readme/badge/meshtastic/python)](https://cla-assistant.io/meshtastic/python)
 [![Fiscal Contributors](https://opencollective.com/meshtastic/tiers/badge.svg?label=Fiscal%20Contributors&color=deeppink)](https://opencollective.com/meshtastic/)
+![GPL-3.0](https://img.shields.io/badge/License-LGPL%20v3-blue.svg)
```
ianmcorvidae, line 8: It looks like this is showing an LGPL badge, but presumably it should be https://img.shields.io/badge/License-GPL%20v3-blue.svg instead?

.reuse/dep5
```diff
@@ -0,0 +1,64 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
```
ianmcorvidae, line 1: This file seems fine to me, but I'm concerned about it falling out of date. Was this file generated by some mechanism, and if so, is there a way we can wire it into our GitHub Actions to ensure this file remains up-to-date?

ianmcorvidae (changes requested):
Sorry to be so slow to get back to this; I have a couple changes & questions below. I do also think the notes on each file will be hard to maintain, in part because many of those files are generated from the protocol buffers (all the `_pb2.py` and `_pb2.pyi` files) and so their headers would need to be integrated by a protobuf plugin or similar. Looking at reuse.software, they mention the `dep5` file as an acceptable alternative, so as long as we can ensure that file stays up to date by some means, that might be fine.

I think technically this counts as a license change, since the prior LICENSE.txt has stuck around, but Apache (or Apache/GPL3 dual license -- it's a little unclear) to GPL3 shouldn't be a problematic relicensing to my knowledge. I would appreciate anyone with more experience in this regard weighing in on the consequences of affirmatively stating this library is fully GPL3 at this point, however.

README.md
FedericoCeratto: Good spot, fixed.

.reuse/dep5
FedericoCeratto: The file is handwritten starting from a template from a hacky script. It's not uncommon for such files to be a bit out of date and sometimes authors don't bother with putting their full name and email address, but this does not limit or voids author's rights: copyright law applies automatically and the file is a reminder for end users, a way to help legal teams check authorship, and a way to recognize people's efforts.

FedericoCeratto: FWIW `reuse lint` can be run automatically in CI and authors can be suggested to update the file when contributing. Some projects use `AUTHORS.md` or `CONTRIBUTORS.md` similarly.

FedericoCeratto:
I removed the second commit adding headers to every file.

With this license clarification if any author contributed code assuming it was under Apache-2.0 license, they can still use those portions of code under Apache license if they want to. (In general, copyright changes cannot take away rights from authors). The change makes the licensing clear to authors and users for future contributions and releases.

garthvh:
> I removed the second commit adding headers to every file.
> 
> With this license clarification if any author contributed code assuming it was under Apache-2.0 license, they can still use those portions of code under Apache license if they want to. (In general, copyright changes cannot take away rights from authors). The change makes the licensing clear to authors and users for future contributions and releases.

Is this really a necessary distinction to make based on this? https://www.apache.org/licenses/GPL-compatibility.html

ianmcorvidae:
I don't think we need to explicitly state the distinction; it remains true, regardless. I think this looks good and I'll go ahead and merge it. I'll look at that `reuse` tool and will also make an issue for potentially running it in CI.

njh:
What was the reason for putting the GPL 3.0 license in `LICENSES/GPL-3.0-only.txt`, rather than in `LICENSE.txt`?

GitHub non-longer displays the name of the detected license with the repository information.

FedericoCeratto:
@njh https://reuse.software/ introduced a machine-parsable format that can describe all the licenses used in a project and automatically track licensing across projects.

PR #584, opened 2024-06-01
Use new fixed position admin messages and add `--remove-position` argument

### LINKED ISSUE #525: [Bug]: setting fixed lat/lon/alt now REQUIRES ``--set position.fixed_position true`` even if ALREADY set to True
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.3.1.4fa7f5a

### Description

I used fixed position with previous fw version. after upgrading to this one, I cannot set it via cli nor via android app.

For cli see log output below, for app I set it in gui, reconnect and shows as lat/long/alt of 0.0.

### Relevant log output

```Shell
``
$> meshtastic --setlat 37 --setlon -121 --setalt 30
Connected to radio
Fixing altitude at 30 meters
Fixing latitude at 37.0 degrees
Fixing longitude at -121.0 degrees
Setting device position

$> meshtastic --export-config
...
  position:
    broadcastSmartMinimumDistance: 100
    broadcastSmartMinimumIntervalSecs: 30
    fixedPosition: true
    gpsMode: NOT_PRESENT
    gpsUpdateInterval: 120
    positionBroadcastSecs: 900
    positionFlags: 811
...
```

ianmcorvidae:
Fixes #525

Pending more restructuring of the CLI, of course, but I've put off fixing this particular thing long enough, I think.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -257,33 +257,41 @@ def onConnected(interface):
         if not args.export_config:
             print("Connected to radio")
 
-        if args.setlat or args.setlon or args.setalt:
+        if args.remove_position:
+            if args.dest != BROADCAST_ADDR:
+                print("Setting positions of remote nodes is not supported.")
+                return
+            closeNow = True
+            print("Removing fixed position and disabling fixed position setting")
+            interface.localNode.removeFixedPosition()
+        elif args.setlat or args.setlon or args.setalt:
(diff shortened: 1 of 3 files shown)


ianmcorvidae:
Right, the other thing I did was allow setlat/setlon to take integers premultiplied by 1e7. Should allow a little more flexibility, hopefully, but willing to remove if we think it'll be problematic.

Issue #125, opened 2021-11-30
static code analysis tool(s)?


mkinney:
Should we add some static code analysis tools?

Some good options are:

https://pylint.org/ 

and

https://github.com/psf/black

Why should we add some static code analysis?
- It helps us find potential issues before they become bugs that our users find.
- It helps to have a more consistent, stable codebase

PR #586, opened 2024-06-03
Reformat code using black 24.4.2 with -t py39

FedericoCeratto:
Hello and thanks for developing this tool.
In the past the use of Black to format the code was proposed in https://github.com/meshtastic/python/issues/125
I would like to propose this PR to reformat the project using Black 24.4.2 with `-t py39` and no other option.
A next step, if there is interest, could be to document the use of Black and recommend running it before commits.


ianmcorvidae:
I know that `black` is being used fairly broadly in Python these days, but I'm not super familiar with the different presets and options it provides. If you're able to, could you explain the choice of `-t py39`?

Also, ff we're going to do this, I would also definitely like it to end up in the GitHub Actions as part of CI in whatever fashion is appropriate for the tool (perhaps simply a check that the code is unchanged by running the tool, but there may be preferred methods). Do you have any insight on that?

Issue #579, opened 2024-05-18
Discussion: refactoring/rearchitecting to separate concerns


ianmcorvidae:
The current codebase has a lot of different things interleaved in often unpredictable ways; for example, stuff that controls how data is printed to the console being in the same basic areas of the code as sending and receiving data, and both of those together with code that controls things like managing client connections.

We should more clearly separate these things for a couple reasons:
 * those using the library to write larger applications that do things very differently from a CLI shouldn't be subject to how data is printed to a console, but should be able to reuse convenience functions around things like administration, sending different sorts of data formatted correctly, etc.
 * those writing separate special-purpose CLI scripts can benefit from some parts of our CLI code (for example, setting up an appropriate client from host/port/BLE options, or waiting for acknowledgements) but not need/want everything we do right now.
 * it's a lot easier to maintain something where it's easier to find each bit of code!

I think this might get broken up a few different ways and I'm not sure exactly what I think should be done here, but I wanted to get the issue out here to discuss things.

Closely related to #569 since separating the concern of presentation from that of acquiring data is key to that.

Previously: #566

holdenweb:
I would suggest we consider creating/consolidating a set of lower-level primitives from the existing wireless driver code and look at embedding this tried-and-tested functionality in a brand new framework created to a more usable command-line environment, using cleaner internal data structures and (I suspect) greatly simplifying much of the logic.

The first step will therefore be to define the radio API (my familiarity with the existing software isn't yet good enough for me to confidently say what parts currently exist) for the layers above to use. There's some milage in that discussion.

ianmcorvidae:
So, several months later, but I think I have a bit of a clearer idea of what can happen here.

Our most core/fundamental interaction with the radio is, of course, FromRadio and ToRadio packets. `MeshInterface._handleFromRadio` basically is in charge of the former and is called by the various subclasses when they receive FromRadio packets, while the latter is handled by `MeshInterface._sendToRadio`, which calls out to `_sendToRadioImpl`, which is expected to be implemented by each connection type.

[...]

PR #587, opened 2024-06-03
Replace timeago

FedericoCeratto:
Replace the timeago library with a simple function. This reduces the amount of Python dependencies making the tool easier to package for Linux distributions. Additionally it handles unexpected timestamps from the future, if any, by returning `None`

### DIFF
--- a/meshtastic/__init__.py
+++ b/meshtastic/__init__.py
@@ -76,7 +76,7 @@ from typing import *
 
 import google.protobuf.json_format
 import serial # type: ignore[import-untyped]
-import timeago # type: ignore[import-untyped]
+from dotmap import DotMap # type: ignore[import-untyped]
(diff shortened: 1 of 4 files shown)


ianmcorvidae:
I think that this largely looks fine to me, and I certainly like getting rid of dependencies and creating more code we can properly add types to (and already have, basically). If you're able to, I do think I would like to see some tests of this code, since by no longer relying on an external library we're probably slightly more responsible for ensuring the output is sane! I suspect this code would also be likely to be moved around by #579, but that is of course still in planning stages, just noting here for good measure.

ianmcorvidae:
I fixed up pylint complaining about a lack of a docstring, and added a test using fuzzing (just for fun, really, I've been learning how to use that library recently) and merged. Thanks for the contribution!

PR #589, opened 2024-06-06
Allow a faster nodedb-less startup on 2.3.11+ with `--no-nodes`

ianmcorvidae:
(using the magic value from meshtastic/firmware#3949 )


thebentern (approved):
No nodedb, no problem! Time to party

ianmcorvidae:
woo, party! though honestly this is pretty nice for a lot of simple stuff with the CLI, will be nice to have it available haha

PR #595, opened 2024-06-14
A small offering of unit test coverage

nerdenator:
Increasing unit test line coverage a bit for some of the BLE classes; very very minor code cleanup where I remove a line of code that simply will not be reached.

### DIFF
--- a/meshtastic/ble_interface.py
+++ b/meshtastic/ble_interface.py
@@ -57,7 +57,6 @@ class BLEInterface(MeshInterface):
         except BLEInterface.BLEError as e:
             self.close()
             our_exit(e.message, 1)
-            return
 
         logging.debug("Mesh init starting")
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
It looks like at present this doesn't add any non-commented-out tests -- I'm guessing you're still working on it? If not, let me know.

nerdenator:
@ianmcorvidae Yes, I apologize, been working with other stuff. Basically, I think I need more isolation around the Bluetooth library; it still seems to be calling code that looks for a Bluetooth device on whatever machine is running the unit tests. This won't fail on my machine (which has Bluetooth) but will fail on GitHub Actions' host node, which if I'm guessing, probably doesn't have Bluetooth.

nerdenator:
@ianmcorvidae I think I'm at a stopping point here; there's some details around the Bluetooth connection code that I need to understand better and maybe dig deeper on. I'll try to make draft PRs in the future.

PR #602, opened 2024-06-20
Improve ACK handling: pass to `onAckNak` and on request in sendData

ianmcorvidae:
correctly pass them to `onAckNak` handlers, and add a mechanism for other handlers to request acks as well.

issue discovered looking at #594; not marking that fully closed though, because there's still an improvement to be made there in having implicit ACKs be ignored if a full ack desired/expected

### DIFF
--- a/meshtastic/mesh_interface.py
+++ b/meshtastic/mesh_interface.py
@@ -296,6 +296,7 @@ class MeshInterface: # pylint: disable=R0902
         wantAck: bool=False,
         wantResponse: bool=False,
         onResponse: Optional[Callable[[dict], Any]]=None,
+        onResponseAckPermitted: bool=False,
         channelIndex: int=0,
     ):
         """Send a data packet to some other node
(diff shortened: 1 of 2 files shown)


meshtastic/mesh_interface.py
```diff
@@ -1129,16 +1134,18 @@ def _handlePacketFromRadio(self, meshPacket, hack=False):
...
-                        if not isAck or (isAck and handler.__name__ == "onAckNak"):
-                            logging.debug(f"Calling response handler for requestId {requestId}")
-                            handler.callback(asDict)
+                # we keep the responseHandler in dict until we actually call it
+                handler = self.responseHandlers.get(requestId, None)
+                if handler is not None:
+                    if (not isAck) or handler.callback.__name__ == "onAckNak" or handler.ackPermitted:
+                        handler = self.responseHandlers.pop(requestId, None)
+                        logging.debug(f"Calling response handler for requestId {requestId}")
+                        handler.callback(asDict)
```
ianmcorvidae, line 1148: (this is the change that fixes the bug -- previously we were not checking for `onAckNak` before shortcutting out of the handler in line 1135. This instead fetches the handler and checks if it should pass acks to it, but only removes it from the handlers if it decides it should)

meshtastic/mesh_interface.py
```diff
@@ -315,6 +316,10 @@ def sendData(
             onResponse -- A closure of the form funct(packet), that will be
                     called when a response packet arrives (or the transaction
                     is NAKed due to non receipt)
+            onResponseAckPermitted -- should the onResponse callback be called
+                    for regular ACKs (True) or just data responses & NAKs (False)
+                    Note that if the onResponse callback is called 'onAckNak' this
+                    will implicitly be true.
```
ianmcorvidae, line 322: Right now this is only added to sendData, but I will try to go through and add it to other helpers too, for library users' sake.

ianmcorvidae:
I'm not sure why codecov is complaining, but everything else seems fine, so I'll go ahead and merge anyway I think.

PR #605, opened 2024-06-21
Add a whitelist of known meshtastic USB VIDs to use a default serial …

geeksville:
Some VIDs are known meshtastic mfgs.  If there are multiple USB ports seen but only one of them is a meshtastic vendor, default to that device.  Old behavior was to show an error asking user to pick.  Users can still manually pick with this new behavior.

This nicely helps when device USB device #s change and means many times users don't need the --port option.

Initially only RAK4631 and heltec tracker are listed


ianmcorvidae (approved):
Looks straightforward!

Issue #604, opened 2024-06-21
Switch from raw virtualenvs for building to Poetry


geeksville:
[Poetry](https://python-poetry.org/) is a IMO very nice/modern/easy/repeatable python build/deployment/release tool.  It is pretty popular and I wish I had used it originally.   It also handles releasing to pypi, multiple build/execution flavors and replaces requirements.txt.

Underneath it is still virtualenv based, so any dev used to the current virtualenv workflows can still build/work that way if they wish.

I had chatted here with folks about a month ago and the idea seemed generally well received.  So I have a working version in my python [dev tree](https://github.com/geeksville/Meshtastic-python/tree/dev).

There are some remaining items:
* make sure all the existing CIish rules work
* update the dev docs to give basic "how to build using poetry" instructions.  Include a section on "how to make it look like the old virtualenv thing if that's your jam"
* clean up into an atomic PR

Making this issue for our discussion and so I can track my todo list.

ianmcorvidae:
Looks like this didn't self-close itself so handling that now!

PR #606, opened 2024-06-21
Change build to use poetry 

geeksville:
I'm just opening this PR (as draft hopefully), so I can watch all the horrible ways this breaks CI.  I'll rebase and edit (and add docs) before this is ready to submit.

remaining TODOS:

* DONE update with recent requirements.txt changes
* DONE update with latest setup.py and delete setup.py
* DONE split tunnel out as an optional variant
* DONE document how to install/use/poetry
* DONE document how to use "poetry shell" to facilitate debugging
* NOT DONE - SEE BELOW test pypi upload from poetry


geeksville:
Hi ya'll!

So I think this PR is probably good to go, except for three caveats:

* I'm waiting to push the docs to the web group until after this change is in.  You can see the new docs here: https://github.com/meshtastic/meshtastic/pull/1309

* I've only half tested the github release workflow.  I _think_ it is probably good (tested most of it locally on my desktop using the act tool).  When the new python leads (@ianmcorvidae ?) are next ready to do a release they might find (small?) problems.  I'm happy to help then as needed.

* I chose to not use the (built-in) pypi upload support of poetry so as to minimize the changes to the release workflow CI script.  Instead it just builds to dist/ like the old setup.py version.

Thoughts?  Please review kinda carefully.  Poetry means we no longer have a setup.py/requirements.txt or the various crufty scripts I wrote to hand increment version.  

I think if you haven't used poetry before you might really dig it.  It makes everything much more standardized and seems to be the 'modern' way of robustly developing/releasing python packages.

btw: I added a run-ci-local.sh script to bin.  It allows the user to run the github CI actions on your desktop which is super useful for testing this stuff.

geeksville:
fixes #604

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+pyqrcode = "^1.2.1"
+tabulate = "^0.9.0"
+timeago = "^1.0.16"
+webencodings = "^0.5.1"
+requests = "^2.31.0"
+pyparsing = "^3.1.2"
+pyyaml = "^6.0.1"
+pypubsub = "^4.0.3"
+bleak = "^0.21.1"
+mypy = "^1.10.0"
```
njh, line 24: Sorry, I know nothing about Poetry but Is it possible to separate out the things that are needed for testing vs. runtime?

geeksville: yes - there is a separate dev depends.  bleak and pypubsub need to stay in the runtime depends.  But good point mypy can move to the dev depends.  I'm not sure about pyyaml - haven't looked to see what that's used for.

geeksville: heh!   I just grepped and I can't find anything in meshtastic even using pyyaml.  I'll look at the history to see when it was added.

geeksville: pyyaml is used for some sort of save as yaml feature added sometime ago - so it needs to stay in runtime deps.

ianmcorvidae (approved):
Sorry it took me a bit to get to this! Looks good to me, so I'll merge it now. I also just merged the PR removing the need for the `timeago` library, so I'll handle any conflicts there/remove that library from our dependencies after merging.

PR #611, opened 2024-06-25
fix bitrot in an old sanity test - use correct namespace

geeksville:


ianmcorvidae:
Cool, thanks. Haven't been looking at those files, hadn't caught this.

PR #613, opened 2024-06-26
Move protobuf python glue from meshtastic to meshtastic.protobuf

geeksville:
Seems to work nicely.

Two things not checked:
* The CI 'update-protobufs' rule - because that seems to be done only when we want to bump protobuf versions.  But I **think** I've properly changed it so it works with the new locations
* I slightly changed how the hack for OS-X is handled by bin/regen-protobufs.sh (btw this entire script is now quite different due to the extensive sed renaming that needs to be done).  Someone should test on OS-X and if there are problems see the Darwin check in this file.

I'm going to include a third commit that includes a manual update of the glue files (so that update-protobufs' doesn't need to be run immediately.


ianmcorvidae:
Heh, I should have mentioned that we haven't been using the update-protobufs workflow in CI at all recently; I've been manually updating the submodule and then running the regen-protobufs script. Getting the CI version back to usable would be nice, though. I'll give this a look when I can give it some attention, since I want to do a little bit of due diligence with regard to both the new subfolder and the upgrade to nanopb 0.4.8, and probably try updating some of my own scripts and things that use the glue packages just to make sure the upgrade process is fairly smooth.

ianmcorvidae:
I think this looks good, and a lot of external code doesn't even seem to need modification right now (I think because of the ones we're importing in `__init__.py`, so a lot of `from meshtastic import X_pb2` still work fine for now). I'll go ahead and merge it now and I'll put a warning on the next release about it. A good trial run for future refactoring anyway, to see if folks have the information to update their own stuff if needed.

PR #617, opened 2024-06-29
We don't need setuptools anymore (I think?) (poetry mediates all that…

geeksville:
… stuff)


geeksville:
alas pkg_resources apparently requires this so bailing on this idea

ianmcorvidae:
I think that we might be able to get rid of the call to `pkg_resources' now, and then we can remove `types-setuptools`. It was there because `importlib.metadata` is python 3.8+ only, but with the swap to poetry we don't support 3.7 any more anyway.

Of course it's also just a dev-only typestub library, so it's not doing all that much anyway, but it would be nice to clean it up probably.

PR #618, opened 2024-06-30
make specifying an argument to --seriallog optional

geeksville:
thanks @ianmcorvidae for telling me about the cool nargs='?' option!


ianmcorvidae:
Happy to share the knowledge!

PR #619, opened 2024-06-30
Add bin/git-resolve-poetry-lock.sh to resolve merge conficts

geeksville:
Previously I was just doing poetry update to resolve such conflicts, but that's not really ideal because it updates all libs. poetry lock --no-update is better, but even better to base the changes from their version of poetry-lock and then add only my changes as required by pyproject.yaml.

You can use this script if you get a merge conflict with poetry.lock and it will just do the right thing.

btw: in my powermon/structuredlogging branch I have a commit to add support for the "poe" tool which will provide a shortcut for this script.

### DIFF
--- a/bin/git-resolve-poetry-lock.sh
+++ b/bin/git-resolve-poetry-lock.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -e
+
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
Seems useful, cool

PR #625, opened 2024-06-30
Allow connection args without an argument, add argument aliases

ianmcorvidae:
The connection args are common enough having some aliases would be nice. `-s` (for serial, `-h` is taken for help), `-t` (for TCP), and `-b` (for BLE) are aliases for their respective arguments (and long versions `--tcp` and `--serial` have also been added). All three also accept an argument or not and use a default behavior if not provided (BLE and serial attempt to detect, TCP uses localhost for our linux-native folks).

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -1101,20 +1101,24 @@ def addConnectionArgs(parser: argparse.ArgumentParser) -> argparse.ArgumentParse
     outer = parser.add_argument_group('Connection', 'Optional arguments that specify how to connect to a Meshtastic device.')
     group = outer.add_mutually_exclusive_group()
     group.add_argument(
-        "--port",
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
I think the docs need a couple python CLI updates (I don't see mention of `--no-nodes` or `--remove-node`, for example). This one won't break anything if people use what's already documented or anything, though, just a few extra things that can be left off or passed in differently. 

This also isn't in a released version of the CLI yet, of course. I was thinking of maybe trying to put a release out sometime tonight or tomorrow, so I'll try to go through the docs to check for any updates before I do that and send y'all a PR, unless someone else is feeling especially inspired to update those docs, haha

ianmcorvidae:
I put up meshtastic/meshtastic#1322 which is a little hack-and-slash but gets most of this stuff in. Improvements extremely welcome, as I say over there as well :laughing:

PR #607, opened 2024-06-22
PowerMon/StructuredLogging support

geeksville:
Mostly done but I'd like to request review/merge soonish (because I've got too many open branches with  changes and tire of merging ;-) ) - see bottom comment.

Fixes #603

PR #627, opened 2024-07-01
New LogRecord protobuf BLE logging

thebentern:
Kissing cousins ticket for https://github.com/meshtastic/firmware/pull/4220

### DIFF
--- a/meshtastic/ble_interface.py
+++ b/meshtastic/ble_interface.py
@@ -2,6 +2,7 @@
 """
 import asyncio
 import atexit
+import google.protobuf
 import logging
 import struct
 import time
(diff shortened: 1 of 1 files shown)


meshtastic/ble_interface.py
```diff
@@ -82,6 +90,27 @@ def from_num_handler(self, _, b):  # pylint: disable=C0116
         self.should_read = True
 
     async def log_radio_handler(self, _, b):  # pylint: disable=C0116
+        log_record = mesh_pb2.LogRecord()
+        try:
+            log_record.ParseFromString(bytes(b))
+            log_record.message = log_record.message.replace("\n", "")
```
geeksville, line 96: What about instead of this we not include the new line on the sending side?  Looks great otherwise.

geeksville (approved):

ianmcorvidae:
I think that eventually here we should do something like https://github.com/meshtastic/python/pull/607 is doing in that we should publish a pubsub message for log lines. That PR (WIP, of course) only passes along the log line and not the whole `LogRecord`, though.

I also have a question, which is whether any actual firmware releases went out with the now-legacy logging characteristic. We could possibly exclude it if there weren't any, but I haven't been watching that closely.

Other than my question and the suggestion from geeksville about removing newlines on the send side, I think this is good for now, and we can augment it with the pubsub once we've got something on both the BLE and the streaming interface sides.

thebentern:
Sounds good. I'm moving the newline handling to the device

geeksville:
I agree about pubsubish cleanup of the Log system on the device side.
Because it really never should have been in RediectablePrint and has grown
quite a bit.  I especially want to fix the thing Andre mentioned about
logging also working over the TCP endpoint.

I have a PR half done for this.  I was hoping to send it in before I left
town for a two day camping trip.  I'll be back tonight and try to send it
ASAP.

Also yeah: if we haven't done an official release with the initial logging
over BLE feature hopefully we only need to support both variants briefly.
Because it costs an extra 512 byte buffer at the BLE level.

(Sent from a phone - please ignore typos)

PR #629, opened 2024-07-02
Fix for hopLimit param on sendTraceRoute not being respected

ppicazo:
If you use:
```
interface.sendTraceRoute(dest="!3b46b95d", hopLimit=5)
```

The `hopLimit` param specified is not used for the trace route as expected, the lora config `hopLimit` is used. This fix allows trace routes to be sent with a dynamic hop limit as the `sendTraceRoute` method definition would lead you to believe is possible.


ianmcorvidae:
I think this looks good; might be something we can extend to more functions, but for now keeping it on sendData is probably sufficient. It looks like pylint is complaining about something, so we can get that fixed and get this in!

PR #636, opened 2024-07-08
final powermon / power analysis reporting changes

geeksville:
PowerMeasurement/reporting/analytics stuff is now 'done' (though I'll probably send in occasional small PRs to tweak PowerStress and the code in analytics).  I'll be sending in a PR to the docs project soon with a description on how to setup the hardware and run the tests/generate reports.

Do these changes look okay to you?  Mostly this is new stuff in the previously unused analytics subproject.


ianmcorvidae (approved):
I still don't fully understand the power management stuff, but I think this looks good to merge in. I think we may want to split out some more of the powermon related dependencies to be optional -- we're pulling in a lot by way of some of those since merging the last PR. That can happen later though, too.

Let me know when this is ready to merge as far as you're concerned! I see you've pushed a couple things since marking me to review, so I just want to be sure you're good on that.

geeksville:
@ianmcorvidae  thanks!  yep - I think ready to merge!  I've tried to conditionalize the analysis stuff so most of it only gets pulled in if the optional extra poetry "analysis" is used.

PR #653, opened 2024-08-24
Adds handler function for public key, and adds to nodedb display

jp-bennett:


ianmcorvidae (approved):
Seems good to me for getting this in place with the current structure! Thanks for finding that other FromRadio field I missed with my earlier attempt at wiring up the security config.

PR #654, opened 2024-08-24
Show two-way traceroute result with SNR if available

GUVWAF:
Note: this is going towards the 2.5 branch.

If available, this shows the two-way traceroute result with SNR per hop. I left some comments for the validation, as it will likely serve as an example for other clients as well.

Also added a bool `isDest` to `_nodeNumToId()` (which defaults to `True`) to distinguish between the broadcast address, which is always a destination, and an unknown node in a traceroute.


ianmcorvidae:
I think this looks good and I'll get it into 2.5 now so we can squash anything CI might complain about. Thanks for this!

PR #655, opened 2024-08-24
Adds admin session_passkey handling for 2.5 remote admin

jp-bennett:
Tested with a 2.5 node doing remote admin over pki with another 2.5 node, as well as a legacy admin against an older node. For completeness, probably should test this with a 2.4 local node as well.

### DIFF
--- a/meshtastic/__init__.py
+++ b/meshtastic/__init__.py
@@ -199,6 +199,13 @@ def _receiveInfoUpdate(iface, asDict):
         iface._getOrCreateByNum(asDict["from"])["snr"] = asDict.get("rxSnr")
         iface._getOrCreateByNum(asDict["from"])["hopLimit"] = asDict.get("hopLimit")
 
+def _onAdminReceive(iface, asDict):
+    """Special auto parsing for received messages"""
+    logging.debug(f"in _onAdminReceive() asDict:{asDict}")
+    if "decoded" in asDict:
+        if "admin" in asDict["decoded"] and "from" in asDict:
+            adminMessage: admin_pb2.AdminMessage = asDict["decoded"]["admin"]
+            iface._getOrCreateByNum(asDict["from"])["adminSessionPassKey"] = adminMessage["raw"].session_passkey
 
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
This looks good to me!

Question, though -- do we need this for other admin "set" messages as well? `writeChannel`, `setOwner`, `setURL`, `set_ringtone`, `set_canned_message`, and stuff like "exit simulator", "reboot" etc. come to mind, and there might be some other places we should be pulling in the session key as well (e.g. when requesting channels, if `writeChannel` needs the key outbound, stuff like that)

jp-bennett:
> Question, though -- do we need this for other admin "set" messages as well?

Ah, yes. I suppose there is more to remote admin than just the config fields. You're correct, it will need to grab the key from those responses, and set it when sending those messages.

meshtastic/__init__.py
```diff
@@ -238,4 +245,7 @@ def _receiveInfoUpdate(iface, asDict):
     portnums_pb2.PortNum.STORE_FORWARD_APP: KnownProtocol("storeforward", storeforward_pb2.StoreAndForward),
     portnums_pb2.PortNum.NEIGHBORINFO_APP: KnownProtocol("neighborinfo", mesh_pb2.NeighborInfo),
     portnums_pb2.PortNum.MAP_REPORT_APP: KnownProtocol("mapreport", mqtt_pb2.MapReport),
+    portnums_pb2.PortNum.ADMIN_APP: KnownProtocol(
```
ianmcorvidae, line 248: I think this is just replacing it, but we should probably just add the handler to the existing entry for `ADMIN_APP` up on line 228

ianmcorvidae (approved):
Looks fine, I think! We should consolidate those two entries in the `protocols` dict but otherwise I think this is good. We still need to add the fetch session key bit you mentioned in discord, of course. I'll try to give that a look if someone else doesn't beat me to it

meshtastic/__init__.py
jp-bennett: Missed that it was already in there!

PR #656, opened 2024-08-24
More admin

jp-bennett:


ianmcorvidae:
This seems pretty good. I'm thinking it might be good if we put the `requestConfig` calls into the various functions in `node.py` (`setOwner`, to use your example), just because it doesn't really make sense to call those functions without also getting a session key. In there, it'd also be easier to wrap it in an `if` that doesn't bother fetching a session key if one is already present, and it could all get split out into a helper.

tl;dr in node.py maybe something like:

```python
    def ensureSessionKey(self):
        if isinstance(self.nodeNum, int):
            nodeid = self.nodeNum
        elif self.nodeNum.startswith("!"):
            nodeid = int(self.nodeNum[1:],16)
        if self.iface._getOrCreateByNum(nodeid).get("adminSessionPassKey") is None:
            self.requestConfig(admin_pb2.AdminMessage.SESSIONKEY_CONFIG)
```

and then within `setOwner` (and whatever else), just throw in a `self.ensureSessionKey()`.

I guess it's a little more awkward within that function (probably we should make a utility function that takes whatever format of nodeid and spits out the integer, for one) but it should mean only writing that logic once, at least.

meshtastic/__init__.py
```diff
@@ -202,7 +202,7 @@ def _receiveInfoUpdate(iface, asDict):
 def _onAdminReceive(iface, asDict):
     """Special auto parsing for received messages"""
     logging.debug(f"in _onAdminReceive() asDict:{asDict}")
-    if "decoded" in asDict and "from" in asDict and "admin" in asDict["decoded"]
```
ianmcorvidae, line 205: heh, oops, my bad. I'm a professional software engineer, I promise :rofl:

ianmcorvidae:
I think this looks good so I'll merge it. I think there might be a couple missing spots for `ensureSessionKey` calls but I can go through those more quickly in vim than I can on GitHub anyway :laughing:

PR #658, opened 2024-08-25
Add method to be able to send heartbeat

wnagele:
Allows to check if the connection is still alive and working. I need this in a case where other calls are only happening fairly infrequently and a broken connection would only show up much later.


ianmcorvidae:
This looks good to me (we can ignore the code coverage aspect, I think)! I know a few others have been interested in something like this being split out for easy use.

PR #663, opened 2024-08-30
Don't automatically set the time from Python

jp-bennett:
The Python MO is to do as little as possible beyond what the user has intentionally instructed. So don't try to set the time automatically.


thebentern (approved):

ianmcorvidae:
Seems alright to me. I'll merge it and add an extra commit to make `--no-time` a no-op (with just this change it'll still send a position, just not with a time). We can add a `--set-time` that uses the new admin command in 2.5+

PR #664, opened 2024-09-03
Reuse node to prevent overwriting channel settings

lysol:
Hi!

I noticed that I'm unable to use any `--ch-set` settings over the admin channel (using `--dest`). The core issue is that the repeated use of `interface.getNode` fetches all the channels again, overwriting the channels on the node object that had just been retrieved and updated to change the appropriate settings. This means that `downlink_enabled`, `uplink_enabled`, and so on couldn't actually be updated using remote admin. I've confirmed this change allowed me to change MQTT uplink/downlink on a remote node on a tall rooftop.

I attempted to use `requestChannels` in `getNode` but the repeated usage of `getNode` itself blows away the channels on the object, regardless.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -778,7 +778,8 @@ def onConnected(interface):
             channelIndex = mt_config.channel_index
             if channelIndex is None:
                 meshtastic.util.our_exit("Warning: Need to specify '--ch-index'.", 1)
-            ch = interface.getNode(args.dest).channels[channelIndex]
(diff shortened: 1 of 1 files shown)


ianmcorvidae (approved):
This looks great to me. Should also remove a full cycle of requesting channels from these commands, not that they were working before anyway 😅

ianmcorvidae:
(setting CI to run in case it finds anything that needs updating, but other than if it does, ready to merge I'd say)

PR #665, opened 2024-09-04
Retry admin channel setting retrieval and add configurable timeout

lysol:
Couple things I noticed while working with the admin channel with the Python CLI:

* If one request for a channel times out, it's a real bummer, because we have to start over. These changes make it so if there's a timeout, the same request is retried up to a user-configurable number of times that defaults to 3.
* The default timeout for this is 300 seconds, or 5 minutes, which is pretty long. In my experience, if you don't hear back inside of 20 seconds, it's not going to come at all. I made it user-configurable, but maybe dropping it to 30 seconds or something is better, because the changes here are kind of complex for just managing a timeout.
* 
I'm not 100% on the wording of the help messages -- it's not my strong suit. I've tested these changes manually, ~~but didn't make any changes to tests~~ and added (hopefully enough) test coverage. I feel reasonably confident about the changes, but the changes add complexity that wasn't there.

I'll note that in 2.5.x, the admin channel, especially with channel info retrieval, seems to be a lot more reliable, so this might be a lot more complexity that could be unnecessary. If you think that's the case, I think just customizing the timeout might be nice, since five minutes is a pretty long period of time.


meshtastic/__main__.py
```diff
@@ -1423,6 +1429,18 @@ def initParser():
...
+        help=("Attempt to retrieve channel settings for --ch-set this many times before giving up."),
+        default=3,
+    )
+
+    group.add_argument(
+        "--timeout",
+        help="How long to wait for replies",
+        default=300
+    )
+
```
ianmcorvidae, line 1443: I think we can make this code a little nicer by adding `type=int` to both of these arguments, at which point it also shouldn't be necessary to convert them to ints later.

meshtastic/mesh_interface.py
```diff
@@ -314,19 +314,33 @@ def getTimeAgo(ts) -> Optional[str]:
...
+                    retries_left -= 1
+                    if not n.waitForConfig():
+                        new_index: int = len(n.partialChannels)
+                        # each time we get a new channel, reset the counter
+                        if new_index != last_index:
+                            retries_left = requestChannelRetries - 1
+                        if retries_left <= 0:
+                            our_exit(f"Error: Timed out waiting for channels, giving up")
+                        print("Timed out trying to retrieve channel info, retrying")
+                        n.requestChannels(startingIndex=len(n.partialChannels))
```
ianmcorvidae, line 340: I think the CI failures here for the types are correct in that `partialChannels` isn't guaranteed to be a list. I think you could do:

```python
new_index: int = len(n.partialChannels) if n.partialChannels else 0
```
And, you may as well change the `requestChannels` call on line 340 to just reuse the `new_index` variable as well.

meshtastic/node.py
```diff
@@ -25,14 +25,14 @@ class Node:
...
+    def __init__(self, iface, nodeNum, noProto=False, timeout: int = 300):
         """Constructor"""
         self.iface = iface
         self.nodeNum = nodeNum
         self.localConfig = localonly_pb2.LocalConfig()
         self.moduleConfig = localonly_pb2.LocalModuleConfig()
         self.channels = None
-        self._timeout = Timeout(maxSecs=300)
+        self._timeout = Timeout(maxSecs=timeout)
         self.partialChannels = None
```
ianmcorvidae, line 36: I think the CI may expect this line to be set to a type of `Optional[List]`

ianmcorvidae (commented):
I think I like your thinking here in general. Got a couple comments/recommendations to get CI a bit happier, hopefully.

lysol:
Just a note: I ran into a test error with `util.fromStr` involving '0x0' as input, so I added a fix for that. If this is something you'd rather not have in there, I can pull that out, not a big deal.

meshtastic/util.py
```diff
@@ -82,7 +82,7 @@ def fromStr(valstr):
         val = bytes()
     elif valstr.startswith("0x"):
         # if needed convert to string with asBytes.decode('utf-8')
-        val = bytes.fromhex(valstr[2:])
+        val = bytes.fromhex(valstr[2:].zfill(2))
```
lysol, line 85: Added this because `fromStr` failed on `0x0` as fuzzed input. This is unrelated to the other changes, so I can pull this out if necessary.

lysol:
Thanks @ianmcorvidae for the hints on the CI failures -- made it easy to get familiar with the CI stack on this repo.

meshtastic/util.py
ianmcorvidae: Glad to accept some test fixes alongside other stuff, so no worries there for sure!

ianmcorvidae (approved):
I think this looks good to bring in. Thanks for the updates! I know you mentioned in Discord that you've been having fewer issues with 2.5, which is great, but I'm sure this'll still be valuable.

PR #673, opened 2024-09-17
Allow port number to be specified with tcp hostname

djholt:
This is minor but useful when a different port number needs to be specified (when tunneling over SSH, for example).


ianmcorvidae:
Looks like the test failure is unrelated, to do with test fuzzing for some other stuff. This looks good, thanks!

PR #675, opened 2024-09-19
Enable setting and removing fixed position via remote admin

djholt:
Based on recent firmware change to allow remote admin position config.

Question: do we need to handle old firmware versions with this change?


ianmcorvidae:
Looks good! Thanks for picking this up, I hadn't gotten to it yet. I don't think we need to worry about older firmware versions. They'll reject any updates that get sent by this, but if someone hasn't updated they're probably aware of that constraint anyway.

PR #597, opened 2024-06-16
Adding mypy typing

william-stearns:
Initial attempt to add python type hints to the majority of the (non pb2) python files in this repo.  There are a few notes about things that may need attention, but it's largely focused on type hints.

PR #681, opened 2024-10-10
Add a variety of type annotations for mypy

william-stearns:
Replacing #597 (I'm sure there's a more elegant way to do this.)  Fixes a single missing "import io" in stream_interface.py


ianmcorvidae:
Hi @william-stearns -- I went through and fixed the rest of the mypy complaints locally and pushed that up here. Sorry about the back and forth on this and thanks again for contributing! I'll get this merged now that CI is happy.

Issue #678, opened 2024-10-01
configure does not survive PKI keys 


jhps:
In 2.5 (65305af1840f720eb6e3f7aa0772aa0dd6f8c27e), a configuration that has privateKey, publicKey, or adminKeys causes the cli to exit with:
`Aborting due to: expected bytes, str found`

It seems that MessageToDict() converts bytes to base64 strings that are saved in the config yaml but when these are are later loaded into a config to be sent a setattr() of a string to a bytes field causes an exception.

contra-bit:
I'd like to mention that also the Python CLI has an issue with serializing public keys, which always converts them to a string or bytes, regardless of the input format.  This should be fixed to ensure consistent behavior.
   
 Fixes:
 - [ ] Fix the Python CLI issue with serializing public keys.
    
 All of these examples produce the same error. 
 ```
 meshtastic --set security.is_managed true --set security.admin_key b'XXXXXXXXXXXXXXXXXXXXXXXXXX' --set security.admin_channel_enabled true
 meshtastic --set security.is_managed true --set security.admin_key XXXXXXXXXXXXXXXXXXXXXXXXXX --set security.admin_channel_enabled true
 meshtastic --set security.is_managed true --set security.admin_key 'XXXXXXXXXXXXXXXXXXXXXXXXXX' --set security.admin_channel_enabled true
 ```
 In the example with: b'XXXXXXXXXXXXXXXXXXXXXXXXXX' the it tries to add the b to the bytes

The fix should allow document how to add multiple admin keys as per https://github.com/meshtastic/web/issues/315

PR #696, opened 2024-10-18
fix base64 encoding of key field in config

fmoessbauer:
The security.privateKey and security.publicKey fields are of type bytes, but the protobuf MessageToDict converts them to base64 encoded strings. When importing the config again, this is read as a string, which breaks the import. Instead, the value needs to be prefixed with "base64:", so the type handling logic on import kicks in and decodes the value to a bytes array again.

Fixes: #678

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -1034,6 +1034,15 @@ def export_config(interface):
                 prefs[meshtastic.util.snake_to_camel(pref)] = config[pref]
             else:
                 prefs[pref] = config[pref]
+            # mark base64 encoded fields as such
+            if pref == "security":
+                if 'privateKey' in prefs[pref]:
+                    prefs[pref]['privateKey'] = 'base64:' + prefs[pref]['privateKey']
+                if 'publicKey' in prefs[pref]:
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
Cool, thanks for the PR! It might be good if we eventually do this for any `bytes` field, but I think that's a task for another time. Not letting the perfect be the enemy of the good and all that.

fmoessbauer:
> Cool, thanks for the PR! It might be good if we eventually do this for any `bytes` field, but I think that's a task for another time. Not letting the perfect be the enemy of the good and all that.

Agree, I had a look into that as well. The root-cause is the type incorrectness of `google.protobuf.json_format.MessageToDict`. AFAIK there is no way to conserve the type. By that, we can either fix it for all well-known types manually, or we write a heuristic that checks if a string "could be" a base64 value and convert it back (or prefix it with `base64:`, which is BTW also non standard). IMHO this is quite risky as we can have arbitrary user-supplied strings.

A probably better solution is to interface with the protobuf definitions using the class interface instead of passing around plain dicts. This should conserve all types. But this requires a major refactoring.

PR #697, opened 2024-10-19
Add "(MQTT)" in nodes after node name when appropriate

lachesis:
This matches what the Android app does.


ianmcorvidae:
I'm sorry to get to this so slowly. I don't like modifying the name that's returned; while that works for the apps, I think we should retain the values from the node. That being said, once https://github.com/meshtastic/python/pull/736 is merged, it should be possible to tell the CLI to add another column for the via-MQTT flag, which I think is ultimately better. I'm going to close this for now, and hopefully that other PR can be suitable to merge soon.

PR #701, opened 2024-10-25
A lot of fixes around setting / retrieving base64 encoded values

fmoessbauer:
For details, please check the individual commit messages.


ianmcorvidae:
Not sure why the codecov check is failing but I'm not worried about it. This all looks good to me; thanks again for the fixes/improvements!

PR #707, opened 2024-11-24
Allow request-telemetry to illicit LocalStats response

thebentern:


ianmcorvidae:
LGTM. I'm guessing this is a local-only request, or does this work over the mesh? Either way this seems good though, no real need to prevent people trying it in the CLI, haha.

thebentern:
> LGTM. I'm guessing this is a local-only request, or does this work over the mesh? Either way this seems good though, no real need to prevent people trying it in the CLI, haha.

Previously localstats was PhoneAPI only telemtry variant, but in the upcoming 2.5.14 version, it can be requested even over the lora mesh. It just isn't regularly broadcasted on any sort of interval.

PR #710, opened 2024-12-07
Add text message port cli option

digitaldisarray:
## What did you change?
I added a command line option '--textport' that lets you send text to an arbitrary port number.

## Why did you change it?
When developing a new meshtastic firmware module, there isn't a super easy way to send arbitrary text packets to a module. Previously, module developers that just want to receive text data for testing purposes would likely make their module receive normal text messages (port 1) during development which is a bit of a hack-y solution. Or, developers would have to leave the comfort of their command line and write a script to send the messages on the custom port.

## Screenshot
 
screenshot-edited

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -442,7 +442,8 @@ def onConnected(interface):
             channelIndex = mt_config.channel_index or 0
             if checkChannel(interface, channelIndex):
                 print(
-                    f"Sending text message {args.sendtext} to {args.dest} on channelIndex:{channelIndex} {"using PRIVATE_APP port" if args.private else ""}"
+                    f"Sending text message {args.sendtext} to {args.dest} on channelIndex:{channelIndex}"
+                    f" {'using PRIVATE_APP port' if args.private else ''}"
                 )
                 interface.sendText(
                     args.sendtext,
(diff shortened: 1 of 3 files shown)


thebentern:
IMO this should be limited PRIVATE_APP. We don't need to easily allow folks to send arbitrary payloads over common portnums like position and telemetry for instance

digitaldisarray:
That makes sense. I'll modify my PR when I have the time to just be for PRIVATE_APP.

ianmcorvidae:
Agreed on the proposed change. I've set off the CI build in case it brings up anything else that could use improvement, for whenever you get back to it.

digitaldisarray:
Thank you for your patience. I believe the changes I just made limits users to PRIVATE_APP in a reasonable way. Let me know if any other changes should be made. I'll get to them quicker this time :)

ianmcorvidae:
This looks good to me, sorry for the delay in re-checking it. Assuming that CI doesn't complain about anything, I'll get this merged soon.

ianmcorvidae:
Ah, spoke too soon. I guess it's got some issues with syntax, possibly needs some quote-shuffling. I'll try to take a look when I get chance if you don't first.

Issue #692, opened 2024-10-15
Feature:  Add  "ROLE"  to --nodes report 


Paul-Pressure:
Please  report each nodes "ROLE"  to a new column in the --nodes report

migillett:
PR Created: https://github.com/meshtastic/python/pull/730

dandrzejewski:
I'm not seeing any roles appearing in my `--nodes` output because it appears the data isn't coming from the radio.  Is there a minimum firmware required for this, or am I missing something?

migillett:
I was using the latest firmware version when I ran my testing. That, or it's possible none of the nodes in the area were reporting their roles. How many nodes did you have in your DB?

PR #730, opened 2025-01-22
#692 feature/nodes-add-role

migillett:
First contribution to get acquainted with the code base.


migillett:
tested using Windows 11 and a Heltec V3 via TCP. Ran locally using `python .\meshtastic\ --tcp 192.168.3.185 --nodes`. Result printed out as expected with node roles.

migillett:
This is in response to the ticket: https://github.com/meshtastic/python/issues/692

Issue #464, opened 2023-11-16
Accept node.id in decimal and hex


jp-bennett:
To specify a node, we currently use an exclamation mark, which has other effects in bash. It would be ideal to also allow nodes to be specified by their numeric value, as well as by 0xdeadbeef as well as !deadbeef.

dl33t-me:
Would this need to be changed in the firmware or just in the pattern matching from the command line?

ianmcorvidae:
this should only need updating within the CLI/library -- everything is converted to an integer to send to the radio

migillett:
Made a PR to potentially address this issue.
https://github.com/meshtastic/python/pull/731

I'm open to comments on improving this though.

migillett:
PR just merged to main. You can now use `!` or `0x` prefixes when setting node destinations. ie: `--dest "0x12345678"` or `--dest "!12345678"`

ianmcorvidae:
I'll keep this open for the time being, because I think we'd still like to add some handling for passing node IDs in pure decimal at the command line. I think we can add some handling that treats a value as decimal if it has no prefix at all and all digits could be decimal. I'm not sure entirely what we should do with an unprefixed 8-digit all-decimal-digits value, though.

RuffR1d3r:
Happy to pick up the remaining piece of this — pure-decimal node IDs at the CLI.

From the thread so far, the consensus rule would be:

- `!deadbeef` / `0xdeadbeef` → hex node id (already works since #731)
- digits-only, no prefix → decimal node number
- anything else → unchanged behaviour

While implementing this I noticed two things about the current hex branch in `_sendPacket()` (`isinstance(destinationId, str) and len(destinationId) >= 8` → `int(destinationId[-8:], 16)`):

1. digits-only input like `--dest 12345678` currently parses as `0x12345678` (=305419896) — under the rule above it would become decimal instead, so I'd flag that as an intentional change in the PR/docs.
2. any non-hex string of 8+ chars (e.g. a long node name) raises a raw `ValueError` before the NodeDB lookup can run — my branch turns that into the same friendly "not found in DB" exit the rest of the path uses.

On juergenRe's ambiguity point: I checked and destination strings are never resolved against node names today (`self.nodes` is keyed by `!hex` user ids), so there's no name-collision surface. Digit strings shorter than 8 characters currently just error out ("not found in DB"), so decimal support adds a capability; longer ones are the intentional behaviour change flagged in point 1.

[...]

PR #731, opened 2025-01-26
464: allow for 0x node prefix values

migillett:
A potential answer to [issue 464](https://github.com/meshtastic/python/issues/464). This allows users to run commands using a different prefix (`0x`) from the meshtastic CLI.

The code change here checks for if the node ID supplied can be turned into an int or not. If so, just pass the node id directly in. If not, it must be a string. Grab the last 8 characters of the node id and force-format the id to have a `!` prefix. That way users can provide any prefix they want to the meshtastic CLI.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -1649,7 +1649,6 @@ def addRemoteActionArgs(parser: argparse.ArgumentParser) -> argparse.ArgumentPar
         "You need pass the destination ID as argument, like "
         "this: '--traceroute !ba4bf9d0' "
         "Only nodes with a shared channel can be traced.",
-        nargs=1,
         metavar=("!XXXXXXXX", "0xXXXXXXXX"),
     )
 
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
This doesn't look bad, aside from the CI errors (looks like it wants a tuple, not a list, for the `metavar`). However, I'd prefer if this handling was moved into `_sendPacket`, which is what currently processes the `!xxxxxxxx` into integers, and it's a bit lower level. Not everything flows through `sendData`, though most does.

migillett:
@ianmcorvidae ok! I can move this into the _sendPacket function. That's no problem. Do we still want to keep the documentation for [!xxxxxxxx, xxxxxxxxxx] or drop those changes?

ianmcorvidae:
I think those metavar changes are fine but it looks like mypy wants it to be a tuple instead of a list. That should just mean `("!XXXXXXXX", "0xXXXXXXXX")` instead of `["!XXXXXXXX", "0xXXXXXXXX"]` for those, I believe.

I do also notice that it's complaining that `int | str` isn't indexable due to the code you have removing the `!`. With integrating it into `_sendPacket` that may go away, though. If you want to check the stuff that CI does, but locally:

```
poetry run pylint meshtastic examples/ --ignore-patterns ".*_pb2.pyi?$"
poetry run mypy meshtastic/
poetry run pytest
```

migillett:
@ianmcorvidae ok, I think that got it. I tried several things but the metavars deal would constantly throw errors no matter what I did. Once I went back to the old metavars definition of `metavar="!xxxxxxxx"` the tests passed successfully. I instead added a bit to the `help` deal telling about using `!` or `0x` prefixes. Technically no prefixes are required at all with how we're parsing hexadecimal strings with this change though. Want to see what you think about that one.

I also ran a manual test locally between my nodes using `!` and `0x` prefixes and both worked

migillett:
I could also do a `if nodeId.startswith('!') or nodeId.startswith('0x'):` too. but I didn't want to hard-code prefixes unless necessary. But I'm game for either.

migillett:
@ianmcorvidae sorry about that. fixed the issues the linter was yelling about.

ianmcorvidae:
Perfect, no worries. I'll kick it off again in case there's anything else. The linter can get a bit annoying at times so that's understood, haha

ianmcorvidae:
Looks like that's all resolved now. Thanks for the contribution!

PR #736, opened 2025-02-15
Add optional CLI parameter to specify node info fields to show with --nodes parameter.

dandrzejewski:
This PR adds the ability to specify which node information is shown in the table produced by the `--nodes` parameter.  It is optional, and specifying the `--nodes` parameter without the new param outputs the same table it does today.

The new CLI param is `--show-fields` and it requires a comma-separated list of field names. It can only be specified alongside `--nodes`.

For example, `--nodes --show-fields user.id,user.hwModel,position.altitude` will output the following table (the "N" column always shows):

```
╒═════╤═══════════╤═════════════════╤════════════╕
│   N │ ID        │ Hardware        │ Altitude   │
╞═════╪═══════════╪═════════════════╪════════════╡
│   1 │ !4cbc2ae6 │ RAK4631         │ N/A        │
├─────┼───────────┼─────────────────┼────────────┤
│   2 │ !12b5aebe │ RAK4631         │ 382m       │
├─────┼───────────┼─────────────────┼────────────┤
│   3 │ !29dc16d7 │ RAK4631         │ 314m       │
├─────┼───────────┼─────────────────┼────────────┤
│   4 │ !b4640294 │ RAK4631         │ 355m       │
├─────┼───────────┼─────────────────┼────────────┤
│   5 │ !1117fb10 │ TRACKER_T1000_E │ 327m       │
╘═════╧═══════════╧═════════════════╧════════════╛
```

I also did a bit of refactoring in the `showNodes()` function in `mesh_interface.py` related to this.  Rather than simply going through and retrieving individual bits of information out of the data structures, there's now a default list of fields to show and that list contains the fields that are shown today.

There's now a mapping of field names to human-readable names.  Fields that are not in the mapping will be shown with raw field names and values, for example:

`--nodes --show-fields user.id,user.hwModel,position.altitude,deviceMetrics.voltage`

will show the following until someone goes in and adds the name mapping and value formatting.  Note the column header name is not human-readable, the actual values are not formatted to a particular precision, and there's no unit of measure shown.

```
╒═════╤═══════════╤═════════════════╤════════════╤═════════════════════════╕
│   N │ ID        │ Hardware        │ Altitude   │   deviceMetrics.voltage │
╞═════╪═══════════╪═════════════════╪════════════╪═════════════════════════╡
│   1 │ !4cbc2ae6 │ RAK4631         │ N/A        │                   4.098 │
├─────┼───────────┼─────────────────┼────────────┼─────────────────────────┤
│   2 │ !12b5aebe │ RAK4631         │ 382m       │                   3.151 │
├─────┼───────────┼─────────────────┼────────────┼─────────────────────────┤
│   3 │ !29dc16d7 │ RAK4631         │ 314m       │                   2.69  │
├─────┼───────────┼─────────────────┼────────────┼─────────────────────────┤
│   4 │ !b4640294 │ RAK4631         │ 355m       │                   2.917 │
├─────┼───────────┼─────────────────┼────────────┼─────────────────────────┤
│   5 │ !1117fb10 │ TRACKER_T1000_E │ 327m       │                   3.222 │
╘═════╧═══════════╧═════════════════╧════════════╧═════════════════════════╛
```


pdxlocations:
I've given this PR a test drive and functionally it works well.  I'll let a dev comment on the code and structure.  Nice!

meshtastic/mesh_interface.py
```diff
@@ -246,6 +279,29 @@ def getTimeAgo(ts) -> Optional[str]:
...
+            keys = key_path.split(".")
+            value = node_dict
+            for key in keys:
+                if isinstance(value, dict):
+                    value = value.get(key)
+                else:
+                    return None
+            return value
+
+        if showFields is None or showFields.count == 0:
```
ianmcorvidae, line 295: I'm pretty sure that `count` is a method on lists, not a property, so this wouldn't ever be true. I'm guessing you mean `len(showFields) == 0` here.

ianmcorvidae (commented):
Caught one small thing, and also set the CI to run. We'll see if that turns anything up, but otherwise this looks good to me! I'm sure there's some stuff we could do to make it more concise or smart, but I think this way is good for now.

ianmcorvidae:
Ah, yeah, looks like pylint is complaining a bit.

If you want to check the CI stuff locally, you can run:

`poetry run pylint meshtastic examples/ --ignore-patterns ".*_pb2.pyi?$"`

and `poetry run mypy meshtastic/`

and `poetry run pytest`

It looks like I may have messed something up with mypy, though, just pushed a fix for that to master. So if it complains about `argcomplete` then merge in master, or rebase.

dandrzejewski:
> `poetry run pylint meshtastic examples/ --ignore-patterns ".*_pb2.pyi?$"`

Want to run something by you.  Running pylint, I see:

`meshtastic\mesh_interface.py:290: error: Incompatible types in assignment (expression has type "Any | None", variable has type "dict[str, Any]")  [assignment]`

But I'm specifically taking advantage of Python's dynamic typing here to save a line or two.  `value` starts out as a `dict` but becomes a string.

I "fixed" it by indicating that it could be a str, dict or even None:

```python
            value: Optional[Union[str, dict]] = node_dict
```

ianmcorvidae:
Fixed up a couple small things that pylint/pytest were complaining about since I know you were having some install issues from discord. I think this looks good to me, thanks for the contribution!

PR #739, opened 2025-02-19
Added example script : meshtastic_serial_message_reader.py

henri:
Basic example script so you can see messages on a node attached via USB on LINUX. It may work with other connection types and on other operating systems. Your millage will vary. Contributions welcome to improve / expand / extend / modify etc.

### DIFF
--- a/examples/meshtastic_serial_message_reader.py
+++ b/examples/meshtastic_serial_message_reader.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+#
+# Released Under GNU GPLv3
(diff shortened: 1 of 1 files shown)


henri:
Some updates relating to hiding some messages if you want. If there is interest I could add some command line arguments to make it less programmer orientated.

henri:
Tested on MacOS seems to work okay.

henri:
Any issues, feel free to merge in any time. Maybe a file rename would pull this example more in line with the other examples. Guess that can be done later as a tidy up.

PR #747, opened 2025-03-03
Add backup / restore admin commands

thebentern:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "meshtastic"
-version = "2.5.12"
+version = "2.6.0"
```
ianmcorvidae, line 3: I didn't think to update this and should have, but if you're going to here, set it to `2.6.0a1` or something like that. Then the release automation will increment it properly (to `2.6.0` rather than to `2.6.1`).

ianmcorvidae (changes requested):
A couple small things, mostly seems good to me though

pyproject.toml
ianmcorvidae: Heh, failed to include this in the actual review. Well, anyway

meshtastic/__main__.py
```diff
@@ -462,6 +462,21 @@ def onConnected(interface):
             waitForAckNak = True
             interface.getNode(args.dest, False, **getNode_kwargs).removeFavorite(args.remove_favorite_node)
 
+        if args.backup_prefs:
+            closeNow = True
+            waitForAckNak = True
+            interface.getNode(args.dest, False, **getNode_kwargs).re(args.backup_preferences)
```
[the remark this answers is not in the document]
thebentern: Reeeee! Yeah, this one is a mess. I must have been in a hurry

meshtastic/node.py
```diff
@@ -707,8 +707,47 @@ def setFavorite(self, nodeId: Union[int, str]):
...
+        else:
+            onResponse = self.onAckNak
+        return self._sendAdmin(p, onResponse=onResponse)
+
+    def removePreferencesBackups(self, location: Optional[int] = 0):
+        """Tell the node to remove backup preferences from the filesystem."""
+        self.ensureSessionKey()
+
+        p = admin_pb2.AdminMessage()
+        p.remove_backup_preferences = location
```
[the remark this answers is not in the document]
thebentern: Yeah, the only thing that kept me from just using the remove file command is that we do have the SD card stuff as well. I was a bit on the fence

meshtastic/__main__.py
```diff
@@ -1794,12 +1810,40 @@ def addRemoteAdminArgs(parser: argparse.ArgumentParser) -> argparse.ArgumentPars
         action="store_true",
     )
 
+    group.add_argument(
+        "--backup-prefs",
+        help="Tell the destination node to create a backup preferences file."
+        "Location: 0 for local flash, 1 for SD card.",
+        default=admin_pb2.AdminMessage.BackupLocation.FLASH,
+        nargs="?",
+        const=0,
```
ianmcorvidae, line 1819: You probably want `default` to stay `None` (that's what's passed if the argument isn't passed on the command line at all) and to set the `admin_pb2.<...>` value as `const` (which is what's passed if the argument is passed bare, without a real value).

PR #753, opened 2025-03-20
Work around RTS/DTR serial reset issues on Windows

dwhinham:
Some devices (e.g. Heltec V3.1) are reset when the RTS or DTR pins are messed with, which means that without workarounds, Meshtastic CLI will cause a reset with any operation that opens the serial port.

This behavior is documented in this PySerial issue: https://github.com/pyserial/pyserial/issues/124

On Linux, we already handle this by disabling the HUPCL termios flag, which has the effect of preventing the offending lines being toggled. On Windows, setting the initial state of RTS and DTR before opening the port has a similar effect.

Implement this workaround so that Meshtastic CLI can be used on Windows with these devices.

### DIFF
--- a/meshtastic/serial_interface.py
+++ b/meshtastic/serial_interface.py
@@ -46,7 +46,12 @@ class SerialInterface(StreamInterface):
 
         logging.debug(f"Connecting to {self.devPath}")
 
-        # first we need to set the HUPCL so the device will not reboot based on RTS and/or DTR
+        # set port to None to prevent automatically opening
+        self.stream = serial.Serial(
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
Hi, sorry to get to this so slowly. I think this looks good, though I lack a windows machine to test it on, so I'll trust that this is beneficial on that platform for now.

dwhinham:
> Hi, sorry to get to this so slowly. I think this looks good, though I lack a windows machine to test it on, so I'll trust that this is beneficial on that platform for now.

No worries at all - thank you for merging! 🙂

PR #771, opened 2025-05-05
ble_interface: fix disconnected_callback invocation of close

### LINKED ISSUE #770: Potential bug in BLEClient disconnected_callback registration
When investigating why my node sometime ends up in a still-connected state, despite my script crashing (like described [here](https://github.com/meshtastic/python/blob/c3973117c887c8f5854d6dd7ec63d76c35711ab0/meshtastic/ble_interface.py#L81)), I came across this line:

https://github.com/meshtastic/python/blob/c3973117c887c8f5854d6dd7ec63d76c35711ab0/meshtastic/ble_interface.py#L166

context:

```py
        # Bleak docs recommend always doing a scan before connecting (even if we know addr)
        device = self.find_device(address)
        client = BLEClient(device.address, disconnected_callback=lambda _: self.close)
        client.connect()
        client.discover()
        return client
```

Should this be:

```py
client = BLEClient(device.address, disconnected_callback=lambda _: self.close())
```

Note that the lambda actually invokes the method, rather than returning it. According to the documentation, it does seem like the `disconnected_callback` should be a function expecting one argument that will be invoked (rather than to return the handler). So, I suspect this is a bug. Can you confirm?

cc: @thebentern and @geeksville as the last people to touch the line in #616

williballenthin:
closes #770

### DIFF
--- a/meshtastic/ble_interface.py
+++ b/meshtastic/ble_interface.py
@@ -163,7 +163,7 @@ class BLEInterface(MeshInterface):
 
         # Bleak docs recommend always doing a scan before connecting (even if we know addr)
         device = self.find_device(address)
-        client = BLEClient(device.address, disconnected_callback=lambda _: self.close)
+        client = BLEClient(device.address, disconnected_callback=lambda _: self.close())
         client.connect()
         client.discover()
         return client


williballenthin:
I am not able to sign the CLA at this time.

ianmcorvidae:
Sorry about that with the CLA, and thanks for finding the fix. I'll get it handled.

PR #794, opened 2025-06-21
add --set-is-unmessageable to CLI commands

pdxlocations:
allows:
`meshtastic --set-is-unmessagable true`
`meshtastic --set-is-unmessageable true`
`meshtastic --set-is-unmessagable false`
`meshtastic --set-is-unmessageable false`

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -363,11 +363,11 @@ def onConnected(interface):
         if args.set_is_unmessagable:
             closeNow = True
             waitForAckNak = True
-            print(f"Setting is_unmessagable to {args.set_is_unmessageable}")
-            if isinstance(args.set_is_unmessageable, str):
-                val = meshtastic.util.fromStr(args.set_is_unmessageable)
+            print(f"Setting is_unmessagable to {args.set_is_unmessagable}")
+            if isinstance(args.set_is_unmessagable, str):
+                val = meshtastic.util.fromStr(args.set_is_unmessagable)
             else:
-                val = args.set_is_unmessageable
+                val = args.set_is_unmessagable
             interface.getNode(args.dest, **getNode_kwargs).setIsUnmessageable(is_unmessagable=val)
 
         # TODO: add to export-config and configure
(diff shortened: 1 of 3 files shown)


meshtastic/node.py
```diff
@@ -327,6 +327,23 @@ def setOwner(self, long_name: Optional[str]=None, short_name: Optional[str]=None
             onResponse = self.onAckNak
         return self._sendAdmin(p, onResponse=onResponse)
 
+    def setIsUnmessageable(self, is_unmessagable: Optional[bool]=False):
```
ianmcorvidae, line 330: I think we should probably add this to `setOwner`, since we're already messing with all these fields there. And, it should be combined with set-owner and set-owner-short where those are present, which is probably done that way.

meshtastic/__main__.py
```diff
@@ -350,6 +350,26 @@ def onConnected(interface):
                 print(f"Setting device owner short to {args.set_owner_short}")
             interface.getNode(args.dest, False, **getNode_kwargs).setOwner(long_name=args.set_owner, short_name=args.set_owner_short)
 
+        if args.set_is_unmessageable:
```
ianmcorvidae, line 353: Similarly to the comment in node.py, this block should be merged into the `if args.set_owner or args.set_owner_short` above it, so everything's done in one admin command to the node.

ianmcorvidae (commented):
I'm thinking this should be rearranged a little to match other code, but otherwise looks reasonable to me.

meshtastic/node.py
pdxlocations: Yes!  That makes a lot of sense.  Thanks for the review, I'll revise.

meshtastic/__main__.py
```diff
@@ -1540,6 +1555,13 @@ def addConfigArgs(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         "--set-ham", help="Set licensed Ham ID and turn off encryption", action="store"
     )
 
+    group.add_argument(
+        "--set-is-unmessageable", help="Set if a node is messageable or not", action="store"
```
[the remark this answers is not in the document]
pdxlocations: Done!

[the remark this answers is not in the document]
pdxlocations: Hah, thanks for the pointer!  Should be good now.

PR #798, opened 2025-06-30
Allow forced UTF-8 encoding for --export-config

### LINKED ISSUE #728: [BUG] Config export encoding mismatch blocking import
When running --export-config the file generated is using UTF-16 LE encoding, but the --configure option used to import that same file only accepts UTF-8 encoded files.

My environment is python 3.13 running on Windows 11 (not within WSL) using Meshtastic module version 2.5.9.

My current workaround is to use VS Code to save the exported file using the correct UTF8 encoding.

Error during import:
```py -m meshtastic --ble myMACHere --configure .\SND5.yml
INFO file:ble_interface.py scan line:116 Scanning for BLE devices (takes 10 seconds)...
Connected to radio
Aborting due to: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

I'm unable to attach an export of the yml file due to github file restrictions.  I can send the file directly if requested.  I am also adding a link to a Stack Overflow thread that mentions the same issue with Python, but I'm not familiar enough with this module to make the changes myself.

https://stackoverflow.com/questions/3597480/how-to-make-python-3-print-utf8

pdxlocations:
This PR changes the --export-config command to save a UTF-8 encoded file instead of printing to stdout.  Allows the legacy use of > to redirect to a file.

To fix incorrect encoding on Windows Powershell:
```bash
meshtastic --export-config file.yaml
```
Also allow:
```bash
meshtastic --export-config > file.yaml
```

closes #728

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -1468,7 +1468,7 @@ def addImportExportArgs(parser: argparse.ArgumentParser) -> argparse.ArgumentPar
         help="Specify a path to a yaml(.yml) file containing the desired settings for the connected device.",
         action="append",
     )
-    parser.add_argument(
+    group.add_argument(
         "--export-config",
         nargs="?",
         const="-",  # default to "-" if no value provided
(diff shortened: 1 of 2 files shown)


ianmcorvidae (approved):
LGTM, sorry for the delays on my end as always

PR #803, opened 2025-07-09
Async rework WIP

dangerdyke:
Refactor of the interface code to work better with async code in library consumers, starting with separating the connection logic to a new class, as discussed with @ianmcorvidae on discord.

At this moment, this is only a first draft for the new `MeshConnection` class.

### DIFF
--- a/meshtastic/connection.py
+++ b/meshtastic/connection.py
@@ -9,6 +9,7 @@ from typing import *
 
 import serial
 import serial_asyncio
+from bleak import BleakClient, BLEDevice, BleakScanner
 
 from meshtastic.protobuf.mesh_pb2 import FromRadio, ToRadio
 
@@ -16,6 +17,12 @@ from meshtastic.protobuf.mesh_pb2 import FromRadio, ToRadio
 STREAM_HEADER_MAGIC: bytes = b"\x94\xc3"  # magic number used in streaming client headers
 DEFAULT_BAUDRATE: int = 115200
 DEFAULT_TCP_PORT: int = 4403
+BLE_SERVICE_UUID: str = "6ba1b218-15a8-461f-9fa8-5dcae273eafd"
+BLE_TORADIO_UUID: str = "f75c76d2-129e-4dad-a1dd-7866124401e7"
+BLE_FROMRADIO_UUID: str = "2c55e69e-4993-11ed-b878-0242ac120002"
+BLE_FROMNUM_UUID: str = "ed9da18c-a800-4f66-a670-aa7547e34453"
+BLE_LEGACY_LOGRADIO_UUID: str = "6c6fd238-78fa-436b-aacf-15c5be1ef2e2"
+BLE_LOGRADIO_UUID: str = "5a3d6e49-06e6-4423-9944-e9de8cdf9547"
 
 
 class RadioConnectionError(Exception):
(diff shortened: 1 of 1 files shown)


meshtastic/connection.py
```diff
@@ -0,0 +1,180 @@
...
+
+    async def recv(self) -> FromRadio:
+        """Recieve something from the connected device."""
+        async with self._recv_lock:
+            msg_bytes: buffer = await self._recv_bytes()
+            return FromRadio.FromString(str(msg_bytes, errors="ignore"))
+
+    async def listen(self) -> AsyncGenerator[FromRadio]:
+        while not self.on_disconnect.is_set():
+            yield await self.recv()
```
ianmcorvidae, line 84: How do you anticipate this being used in the interface class that would wrap this? Obviously this plus `send` is basically the core API here. In the existing API, we're of course not really expecting callers to set up threads or such themselves, just set up listeners and make an interface, and otherwise go about their business. Part of why I ask is that we also do various management like updating the in-memory node data, and presumably we'd still want that to happen. Would this then only really be async-iterated-over by the interface class, or would it be used by clients directly?

Either way we should probably document it, anyway.

meshtastic/connection.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class SerialConnection(StreamConnection):
+    def __init__(self, portaddr: str, baudrate: int=115200):
+        self.port: str = portaddr
+        self.baudrate: int = baudrate
+        super().__init__(portaddr)
+
+    async def _initialize(self):
+        self._reader, self._writer = await serial_asyncio.open_serial_connectio(
```
ianmcorvidae, line 172: minor, but I'm guessing this is supposed to be `open_serial_connection`

meshtastic/connection.py
```diff
@@ -0,0 +1,180 @@
...
+            port=self._port, baudrate=self._baudrate,
+        )
+
+    @staticmethod
+    async def get_available() -> AsyncGenerator[str]:
+        for port in serial.tools.list_ports.comports():
+            if port.hwid != "n/a":
+                yield port.device
```
ianmcorvidae, line 180: presuming this is planned to eventually flesh out to something similar to what we do in `meshtastic.util.findPorts` and this is effectively a stub for now?

ianmcorvidae (commented):
Obviously not complete yet, but basically looks reasonable to me. I know you have further work planned around reworking the way the async code works, which I assume touches more than just the connections, so some of it will fall into how that integration works.

Which is all a lot of words for "looks good, keep going" I guess, haha

meshtastic/connection.py
dangerdyke: I figured the listen function would be used by the MeshInterface to grab every mesh message, and dispatch pubsub events and such appropriately. I haven't quite worked out the specifics but you're right that it would be problematic to have more then one listen call per connection instance, as implemented.

I think what i'm gonna do is just put a lock on the listen method. Clients using the MeshInterface API won't need to even touch Connection instances directly anyways. I can also document everything before everything's ready to be merged.

meshtastic/connection.py
dangerdyke: Basically, yes. I think it's more sensible to define this behavior alongside the rest of the connection code. I also intend to use this method to implement bluetooth scanning when i get to the BLE connection impl

meshtastic/connection.py
dangerdyke: yep, good catch

PR #810, opened 2025-08-04
Add timeouts to get_ringtone and get_canned_messages

pdxlocations:
When firmware is built without these modules these requests just hang.  This PR adds a 2 second delay before moving on.

### DIFF
--- a/meshtastic/node.py
+++ b/meshtastic/node.py
@@ -448,7 +448,12 @@ class Node:
             self._sendAdmin(
                 p1, wantResponse=True, onResponse=self.onResponseRequestRingtone
             )
-            while self.gotResponse is False:
+            start_time = time.time()
+            timeout = 2
+            while not self.gotResponse:
+                if time.time() - start_time > timeout:
+                    logging.warning("Timeout waiting for ringtone plugin response")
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
Seems reasonable enough for now to get things working again for those cases. A 2s limits feels like it'd be a bit low for doing this over remote admin is my main concern. Would it be possible to move this handling into the export-config part, rather than within the node.py functions? A timeout isn't really a big deal with the individual commands the way it is with export, so I think it'd be fine if only that side has the extra handling.

pdxlocations:
> A timeout isn't really a big deal with the individual commands the way it is with export, so I think it'd be fine if only that side has the extra handling.

In my (limited) testing --get-ringtone and --get-canned-messages would just hang with no indication of what was going on if the modules aren't installed, so maybe it's useful to keep the timeout and log message in node?  

I've not tested with remote admin at all.  What do you think a good delay before timeout would be?

pdxlocations:
Digging in a little deeper I find we have access to `self.iface.metadata.excluded_modules` bitmask.  We could check which modules to exclude when calling these functions.

PR #834, opened 2025-10-05
CLI: normalize lora.ignore_incoming IDs (dec/!hex/0x), dedupe, YAML [] clear, fix bytes→int crash

### LINKED ISSUE #833: [Bug]: lora.ignore_incoming crash (bytes→int), duplicates, and ID format inconsistency
# Config
Python: 3.10
meshtastic-python: 2.7.3
Radios, 2x Heltec V3s, firmware_version: 2.7.10.94d4bdf05

## Description

While managing the LoRa ignore list from the CLI:

meshtastic --set lora.ignore_incoming <id> can crash with Aborting due to: 'bytes' object cannot be interpreted as an integer.

Adding the same node ID multiple times can result in duplicates in lora.ignore_incoming.

Node IDs aren’t consistently accepted across decimal, !hex, and 0xhex formats.

# Reproduction
## Example IDs
ID_DEC=2086287328
ID_HEX_BANG=!7c5a37e0
ID_HEX_0X=0x7c5a37e0

## 1) Crash / empty add
meshtastic --port /dev/cu.usbserial-4 --set lora.ignore_incoming "$ID_DEC"

Sometimes prints: "Adding '' to the ignore_incoming list"
Then: Aborting due to: 'bytes' object cannot be interpreted as an integer

## 2) Duplicates
meshtastic --port /dev/cu.usbserial-4 --set lora.ignore_incoming "$ID_DEC"
meshtastic --port /dev/cu.usbserial-4 --set lora.ignore_incoming "$ID_DEC"
meshtastic --port /dev/cu.usbserial-4 --get lora.ignore_incoming  

This will then sometimes show repeated entries

NillRudd:
Fixes #833 

## Summary
- Accept node IDs in decimal, `!hex`, and `0xhex`; normalize to int.
- Deduplicate `lora.ignore_incoming` on write.
- Support clearing via `config.lora.ignore_incoming: []` reliably.
- Prevent crash: `'bytes' object cannot be interpreted as an integer`.
- Removed some duplicate code.

## Why
CLI operations around ignore list were brittle:
- Occasional crash when adding IDs.
- Duplicate entries possible.
- Inconsistent ID formats accepted.
- Clearing via YAML sometimes confusing.

## Behavior
- `--set lora.ignore_incoming <id>` accepts dec/!hex/0xhex and stores one normalized entry.
- `--get lora.ignore_incoming` reflects deduped list.
- YAML `config.lora.ignore_incoming: []` clears list.

## Notes
Firmware still shows packets in logs; enforcement is firmware-level and out of scope here.
No breaking changes—CLI becomes more permissive and robust.

### DIFF
--- a/meshtastic/node.py
+++ b/meshtastic/node.py
@@ -16,6 +16,7 @@ from meshtastic.util import (
     pskToString,
     stripnl,
     message_to_json,
+    to_node_num,
 )
 
 logger = logging.getLogger(__name__)
@@ -51,20 +52,6 @@ class Node:
             r += ", timeout={self._timeout.expireTimeout!r}"
         r += ")"
         return r
-
-    def _to_node_num(self, nodeId: Union[int, str]) -> int:
-        """Normalize node id from int | '!hex' | '0xhex' | 'decimal' to int."""
-        if isinstance(nodeId, int):
-            return nodeId
-        s = str(nodeId).strip()
-        if s.startswith("!"):
-            s = s[1:]
-        if s.lower().startswith("0x"):
-            return int(s, 16)
-        try:
-            return int(s, 10)
-        except ValueError:
-            return int(s, 16)
 
     def module_available(self, excluded_bit: int) -> bool:
         """Check DeviceMetadata.excluded_modules to see if a module is available."""
@@ -728,7 +715,7 @@ class Node:
     def removeNode(self, nodeId: Union[int, str]):
         """Tell the node to remove a specific node by ID"""
         self.ensureSessionKey()
-        nodeId = self._to_node_num(nodeId)
+        nodeId = to_node_num(nodeId)
 
         p = admin_pb2.AdminMessage()
         p.remove_by_nodenum = nodeId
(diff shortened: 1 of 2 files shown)


meshtastic/node.py
```diff
@@ -52,6 +52,20 @@ def __repr__(self):
...
+        s = str(nodeId).strip()
+        if s.startswith("!"):
+            s = s[1:]
+        if s.lower().startswith("0x"):
+            return int(s, 16)
+        try:
+            return int(s, 10)
+        except ValueError:
+            return int(s, 16)
+
```
ianmcorvidae: I like this structure, but I think this function belongs in `meshtastic.util`, perhaps something like `to_node_num` there. (side note that most of the functions there are camelcase rather than underscores, but I think we may as well follow pep8 when we can).

Doesn't need to block merging, but if you're able to move this there that'd be appreciated!

ianmcorvidae (commented):
Looks pretty good, one thing I'd like to see moved elsewhere. I'll keep an eye on this and probably merge it soon if you don't get back to it.

meshtastic/node.py
NillRudd: This has now been moved to util! Seems to be a bit of mixture between camelcase and underscores but i kept to_node_num.

PR #839, opened 2025-10-13
Plumb timeout from --timeout through MeshInterface

### LINKED ISSUE #838: mesh_interface's waitForAckNak uses the default Timeout() class timeout, making multiple hop remote administration difficult.
[waitForAckNack](https://github.com/meshtastic/python/blob/7554c03a2628b310d3bd8eab67033a8fc46d90c9/meshtastic/mesh_interface.py#L1022) in mesh_interface uses a default instance of [Timeout()](https://github.com/meshtastic/python/blob/7554c03a2628b310d3bd8eab67033a8fc46d90c9/meshtastic/mesh_interface.py#L122), which defaults to 20 seconds.  It's called directly by [node.py](https://github.com/meshtastic/python/blob/7554c03a2628b310d3bd8eab67033a8fc46d90c9/meshtastic/node.py#L168) and [\_main\_.py](https://github.com/meshtastic/python/blob/7554c03a2628b310d3bd8eab67033a8fc46d90c9/meshtastic/__main__.py#L1081).  

It looks like it's set properly in [node.py](https://github.com/meshtastic/python/blob/master/meshtastic/node.py#L36), and passed to [mesh_interface.py](https://github.com/meshtastic/python/blob/7554c03a2628b310d3bd8eab67033a8fc46d90c9/meshtastic/mesh_interface.py#L385), but ultimately doesn't change the _timeout instance in mesh_interface.

I think it might be the cause of #827 and the cause of my local mesh being unable to do remote node administration over many hops.  Everything works, it's just not waiting long enough to get the reply (ie: it can be seen coming through another node a few seconds after the cli gives up).  I don't have logs because I'm a third party.

horrible-knots:
This should fix #838.  It adds a timeout parameter to the tcp, ble, stream, and serial interfaces, which gets passed down to MeshInterface directly or through StreamInterface.  I tested a short and long timeout through serial and tcp and it honors it now.

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -1364,6 +1364,7 @@ def common():
                     debugOut=logfile,
                     noProto=args.noproto,
                     noNodes=args.no_nodes,
+                    timeout=args.timeout,
                 )
             elif args.host:
                 try:
@@ -1378,6 +1379,7 @@ def common():
                         debugOut=logfile,
                         noProto=args.noproto,
                         noNodes=args.no_nodes,
+                        timeout=args.timeout,
                     )
                 except Exception as ex:
                     meshtastic.util.our_exit(f"Error connecting to {args.host}:{ex}", 1)
@@ -1388,6 +1390,7 @@ def common():
                         debugOut=logfile,
                         noProto=args.noproto,
                         noNodes=args.no_nodes,
+                        timeout=args.timeout,
                     )
                 except FileNotFoundError:
                     # Handle the case where the serial device is not found
@@ -1425,6 +1428,7 @@ def common():
                             debugOut=logfile,
                             noProto=args.noproto,
                             noNodes=args.no_nodes,
+                            timeout=args.timeout,
                         )
                     except Exception as ex:
                         meshtastic.util.our_exit(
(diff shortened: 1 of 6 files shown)


roberthadow:
When is a --timeout command likely to be rolled out to a firmware version and documentation?

horrible-knots:
> When is a --timeout command likely to be rolled out to a firmware version and documentation?

It's not a new feature of the CLI or related to the firmware.  It's the timeout on waiting for a response from a node.  It fixes a bug that made MeshInterface use the default 20 second timeout, making it difficult to administer remote nodes over a few hops.

roberthadow:
So is `--timeout 999` a part of meshtasticd or part of of the firmware?  If the first, I assume that it will part of the meshtasticd release that follows 2.7.3. If the second, a version that follows 2.7.12...

If it's a part of meshtasticd, it would be nice if `--timeout` would be sticky, rather than being a parameter required at each execution of meshtastic. 

It would be nicer still if `--timeout` were part of the device firmware user configuration.  That way, any command requiring a LoRa response, no matter whether from CLI over TCP, over serial or USB, or BLE would wait for a response before failing.

PR #846, opened 2025-10-27
Send request broadcast userinfo

Littleaton:
Summary
Added new actions/arguments to send nodedb.

This should be exactly what the android app does when using exchanging user info option on node details

Motivation
I wanted a way to help troubleshoot new nodes on seeing my node in the mesh, as well as a way to quickly populate / add my node to new nodes nodedb.

Example Usage
Send to everyone: meshtastic --broadcast-userinfo
Send to someone and expect reply back: meshtastic --request-userinfo --dest <nodeid>
Send to someone without reply: meshtastic --send-userinfo --dest <nodeid>

Impact
I am not aware of any issues that adding this would cause but may need further testing on other devices.

### DIFF
--- a/meshtastic/mesh_interface.py
+++ b/meshtastic/mesh_interface.py
@@ -36,7 +36,7 @@ from meshtastic import (
     protocols,
     publishingThread,
 )
-from meshtastic.protobuf import mesh_pb2, portnums_pb2, telemetry_pb2, admin_pb2
+from meshtastic.protobuf import mesh_pb2, portnums_pb2, telemetry_pb2
 from meshtastic.util import (
     Acknowledgment,
     Timeout,


meshtastic/__main__.py
```diff
@@ -1876,6 +1906,27 @@ def addRemoteActionArgs(parser: argparse.ArgumentParser) -> argparse.ArgumentPar
...
+    )
+
+    group.add_argument(
+        "--broadcast-userinfo",
+        help="Broadcast your user information to all nodes in the mesh network.",
+        action="store_true",
+    )
+
+    group.add_argument(
+        "--send-userinfo",
```
ianmcorvidae: Let's have this be available as `--send-user` as well, to match the protobuf name more closely.

```suggestion
        "--send-userinfo", "--send-user",
```

meshtastic/mesh_interface.py
```diff
@@ -483,6 +483,60 @@ def sendAlert(
...
+        user.long_name = node_user.get("longName", "")
+        user.short_name = node_user.get("shortName", "")        
+        user.hw_model = node_user.get("hwModel", 0)
+        user.is_licensed = node_user.get("is_licensed", False)
+        user.role = node_user.get("role", 0)
+        
+        # Handle public key - firmware strips it if node is licensed
+        if "public_key" in node_user and not user.is_licensed:
+            user.public_key = node_user["public_key"]
+
```
ianmcorvidae: The `is_unmessagable` flag should be included here.

```suggestion

    if "isUnmessagable" in node_user:
        user.is_unmessagable = node_user["isUnmessagable"]
```

PR #847, opened 2025-10-28
Container: Add initial container for meshtastic-cli

oliv3r:
Just a quick set of files to enable the build of (tagged) containers. Both alpine and debian containers are available (~200MiB/~1.2GiB) allowing us to use meshtastic cli with a quick docker run, instead of having to build/install stuff locally.

### DIFF
--- a/.github/workflows/container-build.yaml
+++ b/.github/workflows/container-build.yaml
@@ -1,0 +1,77 @@
+name: Create and publish Container image
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*.*.*'
+  pull_request:
+    branches:
+      - master
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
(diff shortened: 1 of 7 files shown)


.github/workflows/container-build.yaml
```diff
@@ -0,0 +1,77 @@
+name: Create and publish Container image
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - 'v*'
```
ianmcorvidae: We don't use the `v` prefix for the tags in this repo -- https://github.com/meshtastic/python/tags

I imagine this should probably just build on any tag, or any three-part numeric thing that looks like a version number.

ianmcorvidae (changes requested):
One small comment, but in general I think this is good. I haven't actually tested the containerfiles myself but they look like they should work fine and hopefully you did, anyway!

ianmcorvidae:
Sorry for taking so long on this. I'm going to give it a good look and maybe harden some of our CI stuff as well, but probably I'll merge this shortly. Thanks for the contribution!

PR #856, opened 2025-11-10
--fs-ls

benallfree:


ianmcorvidae (commented):
Dropping a handful of comments while I have a minute. I definitely agree that if this can be refactored out into a helper file that would probably be good. Lot of stuff going on here and `mesh_interface.py` has already been huge forever.

I'll try to give the actual xmodem logic a good look later but wanted to get you the initial comments first.

benallfree:
Closing for #861

PR #882, opened 2025-12-26
fix(cli): route --reply responses to the correct channel

thatSFguy:
Ensures that automatic replies are sent back on the same channel index the message was received on. Previously, all replies defaulted to the primary channel (0), even if the incoming message arrived on a secondary channel.

Modified the `onReceive` handler to extract the `channel` index from received packets. This ensures `interface.sendText` targets the originating channel rather than always defaulting to the primary channel.


thatSFguy:
Found a new bug.  --reply is listening to all channels instead of the set channel idx, so more changes are needed.  Closing the request.

PR #894, opened 2026-01-25
fix(ble): handle BLEError on init with user-friendly messages

lexfrei:
## Summary

Replace raw Python tracebacks with helpful error messages when BLE and serial operations fail.

**Before:**
```
Traceback (most recent call last):
  File "meshtastic/__main__.py", line 1368, in common
    client = BLEInterface(...)
  ...
meshtastic.ble_interface.BLEInterface.BLEError: No Meshtastic BLE peripheral with identifier...
```

**After:**
```
BLE device not found.

Possible causes:
  - Bluetooth is disabled on the Meshtastic device
  - Device is in deep sleep mode
  - Device is out of range

Try:
  - Press the reset button on your device
  - Run 'meshtastic --ble-scan' to see available devices
```

### BLE Error Handling
Wrap `BLEInterface()` constructor in try/except and provide context-aware error messages for:

- **Device not found** — BLE disabled, sleep mode, out of range
- **Multiple devices found** — need to specify which one
- **Write errors** — pairing PIN required, Linux bluetooth group
- **Read errors** — device disconnected unexpectedly

### Serial Error Handling
Handle `MeshInterface.MeshInterfaceError` for connection timeouts:

- **Connection timed out** — device rebooting, firmware updating, serial interrupted

## Test Plan

- [x] Run `meshtastic --ble nonexistent` — should show "device not found" message
- [x] Test with device in sleep mode — should show helpful suggestions
- [x] Test on Linux without bluetooth group membership
- [x] Test serial connection during device reboot — should show timeout message

### DIFF
--- a/meshtastic/__main__.py
+++ b/meshtastic/__main__.py
@@ -1365,13 +1365,60 @@ def common():
                     print(f"Found: name='{x.name}' address='{x.address}'")
                 meshtastic.util.our_exit("BLE scan finished", 0)
             elif args.ble:
-                client = BLEInterface(
-                    args.ble if args.ble != "any" else None,
-                    debugOut=logfile,
-                    noProto=args.noproto,
-                    noNodes=args.no_nodes,
-                    timeout=args.timeout,
-                )
+                try:
+                    client = BLEInterface(
+                        args.ble if args.ble != "any" else None,
+                        debugOut=logfile,
+                        noProto=args.noproto,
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
Hi, sorry for the delay in looking at this. I've adopted it and added some less brittle info on the BLEError class and branched it off that (and added some tests), and will be merging that in shortly. Thanks for the contribution!

Issue #895, opened 2026-01-28
[Feature Request] Implement ESP32 Unified OTA support


skgsergio:
Until now, I've been using in my ESP32 WiFi nodes a build of the former WifiOTA firmware for ESP32 + p espota.py`. But now that this was superseded by the Unified OTA, the web flasher even flashes the ota partition with the new firmware.

So the `--reboot-ota` no longer works as it expects a different message format, including the file hash, and the `espota.py` seems to no longer work as the protocol is different.

Would it be possible to include the functionality in the python library and CLI so we can flash without the Android app? I have several DIY nodes, and it's a pain having to send the firmware to my phone first.

Reference: https://github.com/meshtastic/Meshtastic-Android/pull/4095

fgimenezm:
The merged PR works ok if the node boots into OTA mode using the same IP it had before booting.
But I've tested it with a node that is using static IP and when the node reboots into OTA mode it defaults to DHCP and gets a different IP, making the OTA process fail.

PR #898, opened 2026-01-31
feat: Add ESP32 WiFi Unified OTA update support

skgsergio:
Added ESP32 WiFi Unified OTA update support.

Did not include BLE support as I didn't manage to connect to a device via BLE. I have a draft (AI-generated, because I know little about BLE and its libraries) for BLE, but I'm not going to commit something I can't review and test properly.

Related to https://github.com/meshtastic/python/issues/895

### DIFF
--- a/meshtastic/tests/test_ota.py
+++ b/meshtastic/tests/test_ota.py
@@ -1,10 +1,11 @@
 """Meshtastic unit tests for ota.py"""
 
 import hashlib
+import logging
 import os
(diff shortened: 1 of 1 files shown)


skgsergio:
That should be it.

Nevermore-42:
Tested this on G2 and it works as expected

skgsergio:
Fixed the CI errors

PR #899, opened 2026-02-05
Refactor the Meshtastic TCP pub/sub example to ensure proper resource cleanup and clearer exception handling.

dim5x:
What changed:

- Wrapped the TCP interface lifecycle in a try / finally block to guarantee that `TCPInterface.close()` is always called.
- Initialized `iface` to None to safely handle failures during interface creation.
- Replaced `sys.exit(1)` with re-raising the exception to preserve the original traceback.
- Simplified imports by importing `TCPInterface` directly.
- Kept the example behavior and public API unchanged.

Why:

- The original example never reached `iface.close()` due to the infinite loop and could leave the TCP connection open when interrupted or when an exception occurred.
- Using finally ensures deterministic cleanup and makes the example safer and easier to reason about, especially for users copying it into their own scripts.
- Re-raising the exception provides better diagnostics while still allowing finally to run.

Behavior impact:

- No functional changes to the demo logic.
- Improved reliability and correctness on errors or interruptions.
- More idiomatic Python resource handling.

### DIFF
--- a/examples/pub_sub_example2.py
+++ b/examples/pub_sub_example2.py
@@ -5,10 +5,9 @@
 import sys
 import time
 
-from pubsub import pub
+from meshtastic.tcp_interface import TCPInterface
 
-import meshtastic
-import meshtastic.tcp_interface
+from pubsub import pub
 
 # simple arg check
 if len(sys.argv) < 2:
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
Thanks for this and sorry for the delay. Looks good to me, so I'll merge it now.

PR #901, opened 2026-02-05
Give TCPInterface reconnect logic on write errors

jerub:
* Moving to socket.sendall() is safer, as sendall will send the entire buffer, while send() would return the number of bytes sent and require being called multiple times if the buffer was full.
 * On exceptions: reconnect to the server.
 * On reconnection: make sure using a lock that there isn't a race between the readers and the writers triggering a reconnect.

### DIFF
--- a/meshtastic/tcp_interface.py
+++ b/meshtastic/tcp_interface.py
@@ -4,6 +4,7 @@
 import contextlib
 import logging
 import socket
+import threading
 import time
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
Hi there, sorry for taking so long to get to this. I'm working on integrating this locally to deal with the merge conflicts and the errant commit of changes to the protobufs submodule. Thanks for the contribution!

PR #908, opened 2026-03-01
Fix: Update repeated field checks to use is_repeated property

pdxlocations:
Update protobuf field-inspection logic to use is_repeated instead of label where needed.

~~I have not tested for backwards compatibility.~~


pdxlocations:
Helper function added to test for is_repeated and fallback to label if not present.
Tested with the current minimum dependency of 4.4.21 and up to 7.34.0.

PR #911, opened 2026-03-19
Fix position overwrite by lower-precision data

### LINKED ISSUE #910: Position overwritten by lower-precision data in _onPositionReceive()
## Bug Description

`_onPositionReceive()` in `meshtastic/__init__.py` (line 183) blindly overwrites a node's position in the node database whenever a `POSITION_APP` packet is received, regardless of precision:

```python
iface._getOrCreateByNum(asDict["from"])["position"] = p
```

This specifically affects the **local node's** position. When other nodes relay our position via map reports at low precision (e.g., `precisionBits: 13`), it overwrites the locally-stored high-precision GPS position (`precisionBits: 32`). Since the local device has the GPS internally, any lower-precision update is always an echo from the mesh, never fresh data.

Remote nodes are not affected in practice — any position they broadcast is their own current state at whatever precision their GPS provides.

## Steps to Reproduce

1. Connect to a Meshtastic device with GPS via the Python library
2. Read `iface.nodes[my_node]["position"]` immediately after connection — correct high-precision position
3. Wait 10-15 seconds for mesh traffic
4. Read position again — it has been overwritten with a lower-precision value from a relayed map report

## Observed Behavior

- Local node position changes from correct value (e.g., `32.9315311`) to incorrect truncated value (e.g., `32.9515008`) 
- Multiple unrelated nodes may show the exact same latitude because they all received the same low-precision map report
- Both `latitudeI` and `latitude` fields are overwritten
- The corruption happens every time, typically within 10-15 seconds of connecting

## Expected Behavior

`_onPositionReceive()` should not overwrite the local node's position with a lower-precision echo from the mesh. The local GPS is authoritative.

## Proposed Fix

Check if the position update is for the local node, and if so, only accept updates with equal or better `precisionBits`:

```python
node = iface._getOrCreateByNum(asDict["from"])
is_local_node = (
    iface.myInfo is not None
    and asDict["from"] == iface.myInfo.my_node_num
)
if is_local_node:
    existing = node.get("position", {})
    existing_precision = existing.get("precisionBits", 0) or 0
    new_precision = p.get("precisionBits", 0) or 0
    if existing_precision == 0 or new_precision >= existing_precision:
        node["position"] = p
else:
    node["position"] = p
```

See PR #911 for the implementation.

## Environment

- Meshtastic Python library: 2.7.8
- Device: LILYGO T-BeamSUPREME (ESP32-S3, SX1262, U-blox M10S GPS)
- Firmware: 2.7.15
- Connected via USB serial
- ~200 nodes in mesh

syn-ack-ai:
## Summary

- `_onPositionReceive()` now checks `precisionBits` before updating the **local node's** position
- Prevents low-precision map report echoes from overwriting the local GPS position
- Remote node positions are still updated normally (any broadcast from them reflects their current state)

## Problem

When other nodes relay our position via map reports at reduced precision (e.g., `precisionBits: 13`), the library blindly overwrites our locally-stored high-precision GPS position (`precisionBits: 32`). This causes:

- Local node position jumping to a nearby but incorrect location
- Multiple nodes appearing at the same truncated coordinates
- Position accuracy degrading from ~1m to ~1km after 10-15 seconds of mesh traffic

This only affects the **local node** — since we have the GPS internally, any lower-precision update is always an echo from the mesh, never fresh data. Remote nodes legitimately send their own positions at whatever precision their GPS provides, so those updates should always be accepted.

## Fix

Added a check in `_onPositionReceive()` that detects if the position update is for the local node (by comparing `asDict["from"]` to `iface.myInfo.my_node_num`). For the local node only, it compares `precisionBits` and rejects updates with lower precision. Remote node positions continue to be updated unconditionally.

## Test plan

- [ ] Connect to a Meshtastic device with GPS via the Python library
- [ ] Read local node position immediately after connection (should be high precision, e.g., 32 bits)
- [ ] Wait for mesh traffic with low-precision map report echoes
- [ ] Verify local node position is NOT overwritten by lower-precision echo
- [ ] Verify remote node positions ARE still updated normally (including lower-precision updates)

Fixes #910

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/meshtastic/__init__.py
+++ b/meshtastic/__init__.py
@@ -179,8 +179,30 @@ def _onPositionReceive(iface, asDict):
             logger.debug(f"p:{p}")
             p = iface._fixupPosition(p)
             logger.debug(f"after fixup p:{p}")
-            # update node DB as needed
-            iface._getOrCreateByNum(asDict["from"])["position"] = p
+            # For the local node, only accept position updates with equal
+            # or better precision. The local GPS is authoritative, and
+            # low-precision echoes from the mesh (e.g., map reports relayed
+            # by other nodes) should not overwrite it.
+            # For remote nodes, always accept the latest position since
+            # any update from them reflects their current state.
+            node = iface._getOrCreateByNum(asDict["from"])
+            is_local_node = (
+                iface.myInfo is not None
+                and asDict["from"] == iface.myInfo.my_node_num
+            )
+            if is_local_node:
+                existing = node.get("position", {})
+                existing_precision = existing.get("precisionBits", 0) or 0
+                new_precision = p.get("precisionBits", 0) or 0
+                if existing_precision == 0 or new_precision >= existing_precision:
+                    node["position"] = p
+                else:
+                    logger.debug(
+                        f"Ignoring low-precision position echo for local node "
+                        f"({new_precision} < {existing_precision})"
+                    )
+            else:
+                node["position"] = p
 
 
 def _onNodeInfoReceive(iface, asDict):


ianmcorvidae:
I think this looks decent, so I'll go ahead and merge it. Thanks for the contribution!

PR #913, opened 2026-04-05
Add textchat.py and replymessage.py examples

SilverBull239:
Adds scripts mentioned in TODO.md:

  - textchat.py — Type a message and it broadcasts to the mesh, incoming messages print to the terminal.
  - replymessage.py — passive listener that auto-replies to any incoming message with the sender's signal stats.

  Both scripts support BLE, TCP, and Serial. Attempted to add typing per Roadmap and follow internal conventions.


ianmcorvidae:
I might make some tweaks to these since I'm doing a bunch of other stuff in parallel right now, but these look good to merge. Sorry about the delay in getting to them and thanks for the contribution!

PR #917, opened 2026-04-17
Update factory reset to use integer for config reset

thebentern:
Alternative fix to #914

### DIFF
--- a/meshtastic/node.py
+++ b/meshtastic/node.py
@@ -731,7 +731,7 @@ class Node:
             p.factory_reset_device = True
             logger.info(f"Telling node to factory reset (full device reset)")
         else:
-            p.factory_reset_config = True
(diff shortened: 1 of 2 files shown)


rummykubd:
Thanks @thebentern this seems like a cleaner solution to issue https://github.com/meshtastic/python/issues/919 (and https://github.com/meshtastic/python/issues/920) especially since it’s aligned with making types stricter throughout the repo.

I'll lean on @ianmcorvidae’s judgement given I’m new here, but Copilot’s recommendations seem valid here - the build check does seem to be failing on the issue it called out ([link](https://github.com/meshtastic/python/actions/runs/24559587721/job/71804602758?pr=917)):
image 

With that added (and build checks running passing), I think this may be okay to merge.

PR #918, opened 2026-04-23
StreamInterface: prevent socket/reader-thread leak on handshake failure

nightjoker7:
## Problem

If `StreamInterface.__init__` calls `connect()` or `waitForConfig()` and either raises (handshake timeout, config error, bad stream), the partially-constructed object propagates the exception up — but the reader thread spawned in `connect()` keeps running and the underlying stream/socket stays open. The caller never received a reference to the half-initialized instance, so they cannot call `close()` themselves.

In a reconnect loop (every Python client retrying after a flaky connection), each failed attempt leaks one socket + one thread. Compounds quickly under load.

This is consistent with the symptom shape in [meshtastic/firmware#10101](https://github.com/meshtastic/firmware/issues/10101) (Station G2 TCP drops with MeshMonitor): the MeshMonitor client retries, retries leak fds, eventually the OS pressure shows up as connection refusal until the leaks time out.

## Fix

Wrap the `__init__` handshake calls in try/except. On failure, call `self.close()` (suppressed inside `contextlib.suppress(Exception)` so a secondary error doesn't shadow the original), then re-raise.

Also handle the case where `close()` is called from the cleanup path before `connect()` started the reader thread — `_rxThread.join()` raises `RuntimeError` when called on a never-started thread.

## Tests added

Four unit tests in `test_stream_interface.py`:

1. `close()` is safe when the reader thread was never started
2. `__init__` cleanup runs when `connect()` raises, original exception propagates
3. `__init__` cleanup runs when `waitForConfig()` raises, original exception propagates
4. A failure in `close()` during cleanup does not shadow the original handshake failure

All passing locally, exercise the previously-uncovered patch lines.

## Diff

`meshtastic/stream_interface.py`: +9 / -2 (the actual fix)
`meshtastic/tests/test_stream_interface.py`: +95 / 0 (the tests)

### DIFF
--- a/meshtastic/stream_interface.py
+++ b/meshtastic/stream_interface.py
@@ -1,5 +1,6 @@
 """Stream Interface base class
 """
+import contextlib
 import io
 import logging
 import threading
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
Thanks for the PR! Merged as part of my current merge spree and checking separately if the copilot comments need addressing.

PR #958, opened 2026-07-17
Half-close TCP connections so the last write is not lost

### LINKED ISSUE #957: TCP config writes (--set etc.) are silently lost on Windows
## Summary

Any one-shot config write over a TCP connection is silently lost on Windows.
`meshtastic --host <host> --set lora.region EU_868` reports success, but the
device never applies or persists it. Reads (`--info`) work fine. Same commands
work on Linux.

The CLI exits 0 and prints "Writing lora configuration to device", so it looks
like it worked.

## Cause

`TCPInterface.close()` tears the connection down abortively immediately after the
admin message is written:

```python
self._wantExit = True
if self.socket is not None:
    with contextlib.suppress(Exception):
        self._socket_shutdown()      # socket.shutdown(socket.SHUT_RDWR)
    with contextlib.suppress(Exception):
        self.socket.close()
```

`writeConfig()` does not wait for an ack, so `close()` runs microseconds after
`sendall()`. Calling `shutdown(SHUT_RDWR)` + `close()` while either side still has
unread data makes the stack send **RST** rather than FIN. The device's FromRadio
stream is still arriving, so there is essentially always unread data.

The platform difference: **Winsock discards data that was already received but not
yet read when an RST arrives**, so the admin message the device had sitting in its
receive buffer is thrown away. Linux hands the buffered bytes to `recv()` before
reporting `ECONNRESET`, which is why this has never shown up there.

Nothing is wrong on the device side; it simply never sees the message.

## Reproduction

Against meshtasticd (Windows build, but any device reachable over TCP will do):

```
meshtastic --host localhost --set lora.region EU_868
# reports success
meshtastic --host localhost --info      # region is still UNSET
```

Device log shows the connection dropping a few ms after the config dump, before
the admin message arrives:

```
[ServerAPI] Config Send Complete millis=15292
[ServerAPI] Lost phone connection            <- ~3ms later, recv() = WSAECONNRESET (10054)
```

## It is a race, and the window is tiny

Adding a delay between `writeConfig()` and `close()` fixes it. Bisecting the
linger, against the same device and command:

| linger after `writeConfig()` | result |
|---|---|
| 0s (what the CLI does) | region `UNSET` - lost |
| 0.05s | region applied |
| 0.2s / 10s | region applied |

50ms is enough. The device polls its socket every 5ms, so it is not slow; the RST
simply beats its next read. That also means it cannot be fixed device-side: you
cannot read faster than the gap between the client's `send()` and its `close()`.

## Fix

Half-close first: `shutdown(SHUT_WR)` sends FIN, which lets the device consume what
we just wrote, then tear down as before. Verified against a real device: with the
patch the same `--set` lands (`lora.region : 1 (US)`), without it the region stays
`UNSET`.

PR follows.

## Affected

Any one-shot TCP write from Windows: `--set`, `--seturl`, `--sendtext`, etc.
Long-lived clients (phone apps, anything that stays connected) are unaffected,
because the write has been read long before the connection closes.

Versions: meshtastic CLI 2.7.3 and 2.7.10, Python 3.12, Windows 11.

caveman99:
Fixes #957.

One-shot config writes over TCP are silently lost on Windows: `--set` reports
success and the device never applies it. Reads are unaffected.

## Cause

`close()` went straight to `shutdown(SHUT_RDWR)` + `close()`. Doing that while
either side still has unread data makes the stack send **RST** rather than FIN,
and the device's FromRadio stream means there is essentially always unread data.

**Winsock discards data that was received but not yet read when an RST arrives**,
so the admin message written moments earlier is thrown away before the device gets
to read it. Linux delivers the buffered bytes before reporting `ECONNRESET`, which
is why this only surfaced on Windows.

`writeConfig()` does not wait for an ack, so `close()` runs microseconds after
`sendall()`.

## Fix

Send FIN first via `shutdown(SHUT_WR)`, so the device can consume what we wrote,
wait briefly for the reader thread to drain and exit, then tear down exactly as
before.

The existing `shutdown(SHUT_RDWR)` is deliberately kept: it unblocks a reader
still sitting in `recv()` so `StreamInterface.close()` can join it. The added wait
is bounded (`GRACEFUL_CLOSE_TIMEOUT`, 0.25s) and returns early when the reader
exits, because a device is not obliged to close just because we half-closed.

## Why not fix it device-side

It is a race with a tiny window, and the device already loses it. Bisecting the
linger between `writeConfig()` and `close()`, same device and command:

| linger | result |
|---|---|
| 0s (current behaviour) | region `UNSET` - lost |
| 0.05s | region applied |
| 0.2s / 10s | region applied |

50ms suffices. meshtasticd polls its socket every 5ms and drains in a loop, so it
is not slow - the RST simply beats its next read. Nothing on the device can read
faster than the gap between the client's `send()` and its `close()`.

## Testing

* `pytest meshtastic/tests/test_tcp_interface.py` - 9 passed.
* Two new unit tests: one asserts the half-close/wait/full-shutdown ordering, one
  asserts `close()` survives a peer that has already vanished. The ordering test
  **fails** if `close()` is reverted to its previous behaviour, so it pins the
  regression rather than just describing it.
* End-to-end against a real device (meshtasticd, Windows 11, Python 3.12), same
  command both times:
  * stock CLI: `lora.region : 0 (UNSET)` - lost
  * patched: `lora.region : 1 (US)` - applied

Behaviour on Linux/macOS is unchanged apart from up to 0.25s of extra latency on
close, and less of it in practice since the wait ends as soon as the reader exits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TCP connection shutdown to close gracefully before forcing termination.
  * Added a brief wait for active reader processes to exit cleanly.
  * Improved resilience when the remote connection is already unavailable, ensuring sockets still close without errors.

* **Tests**
  * Added coverage for graceful shutdown ordering and failure handling.

### DIFF
--- a/meshtastic/tcp_interface.py
+++ b/meshtastic/tcp_interface.py
@@ -11,6 +11,11 @@ from typing import Optional
 from meshtastic.stream_interface import StreamInterface
 
 DEFAULT_TCP_PORT = 4403
+
+# How long close() gives the device to consume what we last wrote before forcing
+# the connection down. See close() for why this exists.
+GRACEFUL_CLOSE_TIMEOUT = 0.25
+
 logger = logging.getLogger(__name__)
 
 
@@ -80,6 +85,18 @@ class TCPInterface(StreamInterface):
         server_address = (self.hostname, self.portNumber)
         self.socket = socket.create_connection(server_address)
 
+    def _wait_for_reader_exit(self, timeout: float) -> None:
+        """Wait briefly for the reader thread to drain and exit after a half-close.
+
+        Returns as soon as it exits, or after timeout: the device is not obliged
+        to close just because we did.
+        """
+        rx = getattr(self, "_rxThread", None)
+        if rx is None or rx is threading.current_thread():
+            return
+        with contextlib.suppress(Exception):
+            rx.join(timeout)
+
     def close(self) -> None:
         """Close a connection to the device."""
         logger.debug("Closing TCP stream")
@@ -87,6 +104,20 @@ class TCPInterface(StreamInterface):
         # Therefore force a shutdown first to unblock reader thread reads.
         self._wantExit = True
         if self.socket is not None:
+            # Half-close first. shutdown(SHUT_WR) sends FIN, which tells the
+            # device we are done writing and lets it consume what we last wrote
+            # before the connection goes away -- typically the admin message a
+            # one-shot command such as `--set` just sent.
+            #
+            # Going straight to shutdown(SHUT_RDWR) + close() while either side
+            # still has unread data makes the stack send RST instead. Winsock
+            # then discards data the peer had already received but not yet read,
+            # so the write is lost; Linux delivers it before reporting
+            # ECONNRESET, which is why this only bites on Windows.
+            with contextlib.suppress(Exception):
+                self.socket.shutdown(socket.SHUT_WR)
+            self._wait_for_reader_exit(GRACEFUL_CLOSE_TIMEOUT)
+
             with contextlib.suppress(
                 Exception
             ):  # Ignore errors in shutdown, because we might have a race with the server
--- a/meshtastic/tests/test_tcp_interface.py
+++ b/meshtastic/tests/test_tcp_interface.py
@@ -1,6 +1,7 @@
 """Meshtastic unit tests for tcp_interface.py"""
 
 import re
+import socket
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


ianmcorvidae:
Looks reasonable to me. I'll pull it in shortly, just seeing if the simradio_testing daily job failure is actually anything or just noise.

caveman99:
It happened twice in a row now.

caveman99:
What actually fails. A single test: test_smokemesh_hop_limit_prevents_relay. It spins up simulated mesh nodes, has A send a broadcast with hopLimit=0, and asserts B receives it directly. B never receives it (assert col_b.wait_for(1) → False). This is firmware-side RF-sim packet delivery — it exercises none of the code your PR touches (the TCP client's graceful half-close). Your change can't cause it and can't fix it.

Why "repeatedly." The daily PPA rebuilds nightly from firmware develop. PR #946's daily leg passed on 2026-07-07; PR #958's fails on 2026-07-17 — a different, later daily firmware. So a firmware change in that 10-day window altered hopLimit=0 direct-delivery behaviour (B no longer receives the direct broadcast), and every re-run of your PR pulls that same broken daily build, so it fails the same test every time. The stable alpha/beta firmware still behaves correctly.

Bottom line: nothing to fix in PR #958 — the daily leg is informational and green-lights the merge regardless. The real signal is a firmware-side regression around hopLimit=0 delivery that belongs in the firmware repo, not here.

PR #965, opened 2026-07-28
Reject malformed tunnel packets

avzamelek:
## Summary

This PR prevents malformed `IP_TUNNEL_APP` payloads from crashing the IP tunnel packet-filtering path.

The tunnel previously assumed every incoming payload was a complete IPv4 packet and indexed packet fields directly. A short, truncated, or non-bytes payload could therefore raise an `IndexError` or formatting/type error while being processed.

## Root cause

`Tunnel._shouldFilterPacket()` reads IPv4 and transport-layer fields at fixed offsets:

- IPv4 protocol byte: offset 9
- Source and destination addresses: offsets 12–19
- ICMP type/code: offsets 20–21
- UDP/TCP source and destination ports: offsets 20–23

Before this change, there was no validation that:

1. the payload was bytes-like;
2. it contained the minimum 20-byte IPv4 header; or
3. ICMP, UDP, and TCP payloads contained enough transport-header bytes for the fields being read.

As a result, an incomplete payload received from the mesh could cause an exception in the tunnel receive path instead of being discarded safely.

## Changes

- Reject non-`bytes`/`bytearray` payloads.
- Reject packets shorter than the 20-byte IPv4 header.
- Reject ICMP, UDP, and TCP packets shorter than the required first four transport-header bytes.
- Log rejected packets at warning level without including packet contents.
- Add regression coverage for:
  - empty payloads;
  - undersized IPv4 headers;
  - non-bytes payloads;
  - truncated ICMP, UDP, and TCP packets.

## Security impact

Mesh payloads should be treated as untrusted input. This change prevents malformed tunnel traffic from triggering an uncaught exception in the packet filtering code, improving resilience against accidental corruption and malformed traffic from remote nodes.

This does not alter handling of valid tunnel packets or change the existing protocol/port blacklist behavior.

## Validation

```text
pytest meshtastic/tests/test_tunnel.py -m unit -q
11 passed, 9 deselected


ianmcorvidae:
This looks pretty reasonable to me. If you can sign the CLA and as long as CI doesn't complain about anything, this is probably good to merge.

Issue #909, opened 2026-03-14
hang with --ble : waiting for <coroutine object BleakClient.write_gatt_char


edsantiago:
Querying my nodes from my Gentoo laptop, meshtastic commands never terminate. They do produce the desired output (info, export-config) but I have to check for the file from another term window, then ^C the meshtastic process.

```shell
$ meshtastic --debug --ble edr2_1e23 --info >info
...does the pause thing...
....then lots and lots of output, as expected, which I'll skip...
...
ground_speed: 1
ground_track: 0
sats_in_view: 7
precision_bits: 32
, 'latitude': 35.xxxxx, 'longitude': -106.xxxxx}}
DEBUG file:client.py write_gatt_char line:816 Write Characteristic f75c76d2-129e-4dad-a1dd-7866124401e7 | /org/bluez/hci0/dev_C3_BF_1E_24_1E_23/service0025/char002b: b' \x01'
DEBUG file:ble_interface.py async_await line:324 async_await: complete
DEBUG file:manager.py _parse_msg line:970 received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_C3_BF_1E_24_1E_23/service0025/char0029): ['org.bluez.GattCharacteristic1', {'Value': <dbus_fast.signature.Variant ('ay', b'')>}, []]
DEBUG file:client.py read_gatt_char line:727 Read Characteristic 2c55e69e-4993-11ed-b878-0242ac120002 | /org/bluez/hci0/dev_C3_BF_1E_24_1E_23/service0025/char0029: bytearray(b'')
DEBUG file:ble_interface.py async_await line:324 async_await: complete
DEBUG file:ble_interface.py async_await line:318 async_await: waiting for <coroutine object BleakClient.disconnect at 0x7f5b6bb17f40>
DEBUG file:client.py disconnect line:396 Disconnecting ({/org/bluez/hci0/dev_C3_BF_1E_24_1E_23})
DEBUG file:manager.py _parse_msg line:970 received D-Bus signal: org.freedesktop.DBus.ObjectManager.InterfacesRemoved (/): ['/org/bluez/hci0/dev_C3_BF_1E_24_1E_23', ['org.bluez.Battery1']]

...about a 2-second pause here.....


DEBUG file:client.py on_data line:876 AcquireNotify: Read error on fd 9: Unexpected EOF on notification file handle. Notifications have been stopped.
[...]

Issue #817, opened 2025-08-22
CLI "-b" and "--ble" options do NOT terminate after operation completed


therickster-cpu:
Using V2.7.1 meshtastic on Linux, doing operations such as

```meshtastic -b Meshtastic_0b88 --ch-index 0 --sendtext "Ping!"```

or

```meshtastic -b Meshtastic_0b88 --no-nodes --info```

sends the message or gets the info, but meshtastic REMAINS connected to the node. It does not terminate (as one would expect it to). The only way out is a CTRL-C. This behavior does NOT occur when using serial connections ("--port" or "--serial" or "-s"). It terminates after the operation as one would expect. I have not tried it via HTTP/HTTPS.

Issue #534, opened 2024-03-31
Linux/Mint: Every meshtastic cli command running just one, two maybe, times


DE8MSH:
Hello.

I can run meschtastic command only one, sometime two, times in same shell as expected.

`meshtastic --host 192.x.x.xx --nodes`

for example. 1st time it runs well. If I want to let run it second (third) time shell shows:

```
meshtastic --host 192.x.x.x --nodes
Error connecting to 192.x.x.x:Timed out waiting for connection completion
```
It seems python / meshtastic doesn't closing any port -- or something like this?!

Anyone?

DE8MSH:
Hmm... Don't ask. But it could have been because Heltec V3 ignored 

`meshtastic --port /dev/ttyUSB0 --set bluetooth.enabled false --set network.wifi_enabled true`

BT was still active. After shutdown BT with Web interface, commands via IP adress working (inmo).

Issue #491, opened 2024-02-25
meshtastic --ble stuck after "Connected to radio"


dirkbeer:
Hi - Really appreciate the BLE support!

I'm having a problem where  `meshtastic --ble` stops after connecting to the radio and nothing further happens:

```
root@cottonmouth:~# . meshtastic/bin/activate
(meshtastic) root@cottonmouth:~# bluetoothctl devices
Device F7:BE:57:1F:90:05 dto_9005
(meshtastic) root@cottonmouth:~# meshtastic --ble-scan
Found: name='dto_9005' address='F7:BE:57:1F:90:05'
BLE scan finished
(meshtastic) root@cottonmouth:~# meshtastic --ble dto_9005
Connected to radio

```

Once I Ctrl-C out of that, `meshtastic --ble-scan` can no longer see the device:

```
(meshtastic) root@cottonmouth:~# meshtastic --ble F7:BE:57:1F:90:05
Connected to radio
^CException ignored in: <module 'threading' from '/usr/lib/python3.11/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.11/threading.py", line 1583, in _shutdown
    lock.acquire()
KeyboardInterrupt:
(meshtastic) root@cottonmouth:~# meshtastic --ble-scan
BLE scan finished
(meshtastic) root@cottonmouth:~# bluetoothctl devices
Device F7:BE:57:1F:90:05 dto_9005
```

Here's my info:

meshtastic --support
```
 System: Linux
   Platform: Linux-6.1.0-rpi8-rpi-v8-aarch64-with-glibc2.36
   Release: 6.1.0-rpi8-rpi-v8
   Machine: aarch64
   Encoding (stdin): utf-8
   Encoding (stdout): utf-8
 meshtastic: v2.2.22
 Executable: /root/meshtastic/bin/meshtastic
 Python: 3.11.2 CPython GCC 12.2.0
```
meshtastic --info
```
Warning: No Meshtastic devices detected.
```

dirkbeer:
Looks like I was just using it wrong, didn't realize that the `--ble dto_9005` has to be included with every command. For example,

`meshtastic --ble dto_9005 --nodes`

works just fine.

ianmcorvidae:
So yeah, the `--ble` bit does need to be included with every command. If you're still seeing it hung after `Connected to radio`, you could try running with `--debug`, which will at least show if it's still attempting to do stuff or not. That might narrow down what's causing it to hang up, potentially.

PR #967, opened 2026-08-07
fix(ble): don't hang on a stalled bleak disconnect during teardown

lalonggone:
## What's wrong

`BLEInterface.close()` can hang forever. I hit this on Linux (Python 3.14, bleak 2.1.1) with a RAK4631: a `--ble` session connects and downloads config fine, then never exits. There are two issues in the teardown path:

1. **Disconnect has no timeout.** `BLEClient.disconnect()` awaits bleak with `future.result(None)`, and `BLEClient.close()` joins the event-loop thread with a plain `join()`. If bleak's disconnect stalls (it does on my setup), `close()` blocks indefinitely and the process wedges.
2. **`UnboundLocalError` on a mid-read disconnect.** In `_receiveFromRadioImpl`, if `read_gatt_char` raises `BleakDBusError` (device dropped), `b` is never assigned, so the following `if not b:` raises `UnboundLocalError: cannot access local variable 'b'` instead of ending the read loop cleanly. This one can bite anyone whose node drops mid-read, I believe on any platform.

## The fix

- Initialize `b` before the read so a mid-read disconnect unwinds through the normal path.
- Add `BLE_DISCONNECT_TIMEOUT` (5s) and apply it to `BLEClient.disconnect()` (swallowing a timeout / `BleakError`, since teardown is best-effort) and to `BLEClient.close()`'s daemon-thread `join()`.

This is deliberately the same "if bleak is hung, don't wait" approach already used on the receive-thread join in `BLEInterface.close()`. I just extended it to the two remaining unbounded waits. The change is additive: if disconnect completes normally the timeout never fires and behavior is unchanged; it only kicks in to stop an otherwise-infinite hang.

## Testing

- `pytest meshtastic/tests/test_ble_interface.py` passes, including two new regression tests (one for the `UnboundLocalError`, one for the disconnect timeout being swallowed).
- `pylint meshtastic/ble_interface.py` -> 10.00/10; `mypy` clean.
- Verified on real hardware (RAK4631) over BLE on both 2.7.15 and 2.7.26 firmware: `--ble --info` and a polling logger now exit cleanly instead of hanging.
- Reproduced across 10 rapid connect / fetch / disconnect cycles: the hang occurred on every cycle and was caught every time, with 0 infinite hangs and 0 unclean exits.

## Related

This looks like the same teardown hang reported in #817 (the CLI not terminating after a `--ble` operation), and may also help #491 and #909 (related Linux `--ble` hangs where the command completes but never exits). Leaving it to a maintainer to confirm before closing any of them.

Possibly related teardown/cleanup reports (not claiming an identical root cause): #49, #534 - both about connections not closing cleanly so later calls fail.

---
*Full disclosure: this is my own bug on my own hardware and I verified the fix here, but I used AI assistance, so I'd really appreciate a couple of extra sets of eyes on this.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth Low Energy disconnect handling with a bounded timeout to prevent stalled shutdowns.
  * BLE timeout and teardown errors are now handled gracefully without interrupting the application.
  * Improved cleanup when a device disconnects during data reception.
  * Enhanced reliability when closing BLE connections and communication threads.
  * Pending operations are now cancelled after communication timeouts, preventing lingering activity.


lalonggone:
Thanks for the review! Addressed both points in 7f44a5f: the disconnect future is now cancelled on timeout, with tests added (assert the bounded timeout is passed, cover the close() join, and cover the cancel path).

The self-join in the disconnect callback is real but pre-existing and separate from this fix, so I've left it out to keep the PR focused and opened #968 to track it.

lalonggone:
Rebased to clean up local commit history, force-push briefly closed this via GitHub's unrelated-history detection. Reopening, no code changes from the last review.

Issue #969, opened 2026-08-10
Bug: Can't run on Termux


sqbi-q:
Importing `serial.tools.list_ports` causes an exception when the platform is unsupported. This stops both CLI and import of Meshtastic as a library. One of those unsupported platforms is Termux (detected as posix). Termux should be able to run Meshtastic as [the documentation provides install instructions for this platform](https://meshtastic.org/docs/software/python/cli/installation/#termux), though.

```
~ $ meshtastic
don't know how to enumerate ttys on this system.
! I you know how the serial ports are named send this information to
! the author of this module:

sys.platform = 'android'
os.name = 'posix'
pySerial version = 3.5

also add the naming scheme of the serial ports and with a bit luck you can get
this module running...
Traceback (most recent call last):
  File "/data/data/com.termux/files/usr/bin/meshtastic", line 3, in <module>
    from meshtastic.__main__ import main
  File "/data/data/com.termux/files/usr/lib/python3.14/site-packages/meshtastic/__init__.py", line 88, in <module>
    from meshtastic.node import Node
  File "/data/data/com.termux/files/usr/lib/python3.14/site-packages/meshtastic/node.py", line 11, in <module>
    from meshtastic.util import (
    ...<10 lines>...
    )
  File "/data/data/com.termux/files/usr/lib/python3.14/site-packages/meshtastic/util.py", line 22, in <module>
    import serial.tools.list_ports # type: ignore[import-untyped]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/data/com.termux/files/usr/lib/python3.14/site-packages/serial/tools/list_ports.py", line 31, in <module>
    from serial.tools.list_ports_posix import comports
  File "/data/data/com.termux/files/usr/lib/python3.14/site-packages/serial/tools/list_ports_posix.py", line 114, in <module>
    raise ImportError("Sorry: no implementation for your platform ('{}') available".format(os.name))
ImportError: Sorry: no implementation for your platform ('posix') available
```

My suggested fix is to wrap import between try-except block:
```py
import serial
try:
    import serial.tools.list_ports
except ImportError as e:
    print("Error: ", e)
    print("Error: Cannot list ports on this platform")
    print("Warning: Skipping import serial.tools.list_ports")
``` 
in `meshtastic/util.py`.

PR #970, opened 2026-08-10
Catch serial.tools.list_ports import exception for unsupported platforms

sqbi-q:
My fix for #969.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application startup when serial port discovery is unavailable.
  * Added clearer diagnostic messages for serial port detection errors and warnings.


ianmcorvidae:
I think the automated review comment is fair, and you'll need to sign the CLA. Other than fixing those things, this seems reasonable as a solution for the import failure.

